Issue #28, opened 2019-06-24
Support all _ definition_ fields to be imported from JSON


pinarpink:
Contact Type, Data Type, Storage Location Category, Document Type, Funding Source. These fields are hardcoded into the initial_data_loader. They should also be imported from JSON.

vildead:
closing issue. Tested on:
develop 309fd6b620425dd4b0de0cefc23851ccb38fd97b

PR #31, opened 2019-07-03
Support all _ definition_ fields to be imported from JSON - issue #28 

kavitarege:


core/management/commands/load_initial_data.py
```diff
@@ -302,19 +256,19 @@ def create_gene_terms():
...
+        # self.create_gene_terms()
+        # self.create_contact_types()
+        # self.create_datatypes()
+        # self.create_sensitivity_classes()
+        # self.create_restriction_classes()
+        # self.create_document_types()
+        # self.create_funding_sources()
+        # self.create_storage_resources()
+        # self.create_roles_and_permissions()
+        # self.create_elu_institutions()
```
neoflex: why are those commented out? is it on purpose?

PR #42, opened 2019-07-18
redirects from http to https

neoflex:
thanks @heinzstockinger  for the suggestion.

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -88,7 +88,7 @@ sudo systemctl restart solr
 
 ## Gunicorn
 
-1) Create the file ```/etc/systemd/system/gunicorn.service``` as the _sudo_ user and with the following content:
+1) Create the file ```/etc/systemd/system/gunicorn.service``` as the _root_ user or with _sudo_ and with the following content:
 
 
 ```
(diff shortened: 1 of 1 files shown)


DEPLOYMENT.md
```diff
@@ -402,6 +401,18 @@ exit
...
     ```
     Changing daisy.com to your particular case.  
+    
+1) To have a redirect from http to https, as _root_ or with _sudo_ create the file ```/etc/nginx/conf.d/daisy.conf``` with the following content:
+
+    ```
+    server {
+      listen 80;
+      server_name daisy.com;
+      return 301 https://daisy.com$request_uri;
```
Trefex: Hi @neoflex,

Using 301 can lead to cache issues when you have a reverse proxy etc in front.

I would recommend using 302 instead.

Thoughts?

neoflex: in this case, it's really a permanent redirect I would say.
If there is a reverse proxy in front of daisy, they should probably do this redirect at the reverse proxy level anyway no? if it's not the case and in the eventuallity that someone would change their mind and now wants to accept http connection, they might have to clear the reverse proxy cache indeed. Using 301 allows taking advantage of its caching by the users' browser.

Trefex: Yes, but clearing that thing from the browser cache is really difficult. But ok, if it's warranted, let's go for 301.

PR #50, opened 2019-07-30
Features/#26 user management

kavitarege:
"DataTables": "^1.10.18"  need to be installed.


neoflex:
for now, it doesn't allow changing passwords and there is apparently no check for password complexity. admin can create, delete and edit users. it's already a nice improvement worth including in the release IMO.

Issue #46, opened 2019-07-30
Permissions not reflected to page behavior


pinarpink:
Two issues with permissions:

* I log in as alice.white (VIP) I am not able to edit dataset sub-entities like storage or access). The view endpoints for these have the ```@permission_required('EDIT', (Dataset, 'pk', 'dataset_pk'))``` annotation.

* I log in as john.doe , he has no rights on the LCSB Human Data Survey v-1 dataset. When he goes to this dataset's overview page, he sees edit and delete buttons.

To reproduce  1. create empty database 2. migrate 3. load initial and demo data

pinarpink:
@neoflex  vcan we add Yohan to the ELIXIR LU organisation? I was not able to assign this issue to him.

yjarosz:
@pinarpink 
* For alice.white issue, I made a pull request that I'll mark for review
* For john.doe, it seems that he is actually `local_custodian` at the project level, so he inherits the rights to edit the dataset.

vildead:
@pinarpink This looks much better! 
One thing: Lately I have created a dummy project and now when I want to access it, I get Server Error (500) even as the admin user. 
Newly created projects or other objects are fine. Might be issue with update procedure? Otherwise we can close this one.

pinarpink:
@vildead  what does the log say for when you get the 500 error?

pinarpink:
@vildead  this should work now after https://github.com/elixir-luxembourg/daisy/commit/425fc268cf69a491767617d0b4bfe5562158c988

vildead:
@pinarpink it works on 425fc268cf69a491767617d0b4bfe5562158c988
I have another small issue:
When the user has admin privileges on a Project and none on one of the datasets in the project, the privileges are cascading - works fine. But when I try to manage permissions of the dataset, I am getting "403 Forbidden". 
Error log:
```
WARNING 2019-08-02 11:38:42,471 log 19069 140614359701312 Forbidden (Permission denied): /permissions/dataset/1/
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 124, in _get_response
[...]

PR #53, opened 2019-07-31
Bugfix/permissions not reflected to page behavior #46

yjarosz:
@pinarpink & @neoflex 
Found it. 

In the permission methods, ones need to use the constants like `Permissions.EDIT` instead of the string `EDIT`.

I found 21 more occurences where it needs to be fixed, like `remove_contact_from_project` or `pick_contact_for_project`....

### DIFF
--- a/core/models/signals.py
+++ b/core/models/signals.py
@@ -1,3 +1,4 @@
+import logging
 from django.conf import settings
 from django.db.models.signals import post_save, m2m_changed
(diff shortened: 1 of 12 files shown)


yjarosz:
Should fix #46

pinarpink:
@yjarosz @neoflex 
Yohan, huge thanks for looking at this.

PR #62, opened 2019-08-01
improve centos update procedure

neoflex:


neoflex:
I've modified the wget commands to specify the output filename because if the files were already present in the folder it was creating files such as hgnc.json.1, etc

Issue #66, opened 2019-08-02
Notification generation on some date fields


pinarpink:
add expiry notification for **DataDeclaration.storage_end_date**

PR #68, opened 2019-08-02
Origin/notification generation on some date fields #66

pinarpink:
@yjarosz @neoflex I added notification generation for nearing storage end dates of data. I will now move onto extending document's with expiry dates. I shall note that the actor+verb+object based notification model does not quite fit  periodic reminders.


core/management/commands/generate_notifications.py
```diff
@@ -0,0 +1,37 @@
...
+            for custodian in ddec.dataset.local_custodians.all():
+                Notification.objects.create(
+                    actor=custodian,
+                    verb=NotificationVerb.data_storage_expiry,
+                    content_object=ddec,
+                )
+
+    def handle(self, *args, **options):
+        self.data_storage_expiry_notifications()
+        print('notifications generated')
```
neoflex: could be interesting to see how many notifications were generated

PR #80, opened 2019-08-07
reorganizing commands in update daisy section

vildead:
The checkout of new version has to be done before downloading fixtures to avoid git complaint about unstaged changes.


neoflex:
we might still have the issue you faced with package-lock.json
maybe we should add `git checkout — web/static/vendor/package-lock.json` to be sure

Issue #65, opened 2019-08-02
Ordinary user can edit contacts/cohorts/partners


vildead:
Should ordinary user (john.doe) be able to edit records like contacts or cohorts as it is now?

yjarosz:
No protection has been st up for those views. I assume it should be protected, right @neoflex ?

yjarosz:
Only data-steward should be able to do?

PR #81, opened 2019-08-08
Ordinary user can edit contacts/contracts/cohorts/partners #65

yjarosz:


yjarosz:
for #65

yjarosz:
So we need to decide which ones can CRUD CCCPs (contacts/contracts/cohorts/partners).
1) Either we give permission only to staff and that simplify things but Users would have to request a CCCP to be added
2) Either we give permission to everyone, but then one should not be able to modify a CCCP added by someone else (unless staff)

Either cases we would need to dev features, request CCCP or Permission mapping. IMHO `2)` is simpler to put in place and we can use `AutoChecker` and `CheckerMixin` to do the job

/cc @pinarpink @neoflex

jLebioda:
It's already deprecated, closing the PR and archiving the branch

PR #84, opened 2019-08-19
adding backup restore steps

vildead:
pg_dump is called with --clean flag but this may cause errors when some tables are missing. That is why I included absolute drop of the database - starting with a clean sheet is easier.
Backup was done by first tar-ing the home directory and then exporting dump into the directory. This should be changed so the .tar archive contains also the database dump, right?


neoflex:
dropping the database first makes sense IMO, yes.
However, the archive should also contain the home directory content and not only the database because the files uploaded to daisy (contracts for instance) but also the log and the configuration need to be back up.

vildead:
thank you, @neoflex 
I have added small updates. 
I have also tested the whole process of backup and restore.

PR #105, opened 2019-11-14
Feature/#44 add export function

pinarpink:
Added export feature with some bare bone  tests.

### DIFF
--- a/core/importer/datasets_exporter.py
+++ b/core/importer/datasets_exporter.py
@@ -1,0 +1,47 @@
+import json
+import sys
+
+from core.models import  Dataset
+from core.utils import DaisyLogger
+from django.conf import settings
+
+logger = DaisyLogger(__name__)
+
+
+class DatasetsExporter:
+
+    def export_to_file(self, file, stop_on_error=False, verbose=False):
+
+        logger.info('Export started ')
+        result = True
+
+        dataset_dicts = []
+        datasets = Dataset.objects.all()
+
+        for dataset in datasets:
+            logger.debug(' * Exporting dataset: "{}"...'.format(dataset.title))
+            try:
+                pd = dataset.to_dict()
+                pd["source"] = settings.SERVER_URL
+                dataset_dicts.append(pd)
+            except Exception as e:
+                logger.error('Export failed')
+                logger.error(str(e))
+                if verbose:
+                    import traceback
+                    ex = traceback.format_exception(*sys.exc_info())
+                    logger.error('\n'.join([e for e in ex]))
+                if stop_on_error:
+                    raise e
+                result = False
+            logger.debug("   ... complete!")
+
+        json.dump({
+            "$schema": "https://git-r3lab.uni.lu/pinar.alper/metadata-tools/raw/master/metadata_tools/resources/elu-dataset.json",
+            "items": dataset_dicts}, file , indent=4)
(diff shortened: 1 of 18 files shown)


core/importer/datasets_exporter.py
```diff
@@ -0,0 +1,51 @@
...
+        logger.info('Dataset export complete see file: {}'.format(file_handle))
+        return result
+
+
+
+    def export_to_buffer(self, buffer, stop_on_error=False, verbose=False):
+        dataset_dicts = []
+        datasets = Dataset.objects.all()
+        for dataset in datasets:
+            logger.debug(' * Exporting dataset: "{}"...'.format(dataset.__str__()))
```
neoflex: minor comment: you can do `see file: {}'.format(dataset)`
even better: `logger.debug(' * Exporting dataset: "%s"...', dataset)` this way, the str method and the concatenation are not executed if not in debug mode

core/importer/datasets_exporter.py
```diff
@@ -0,0 +1,51 @@
...
+                logger.error('Export failed for dataset {}'.format(dataset.title))
+                logger.error(str(e))
+                if verbose:
+                    import traceback
+                    ex = traceback.format_exception(*sys.exc_info())
+                    logger.error('\n'.join([e for e in ex]))
+                if stop_on_error:
+                    raise e
+            logger.debug("   ... complete!")
+        json.dump({
```
neoflex: any reason not to use json.dumps instead and get rid of the buffer and stringio?

core/importer/partners_exporter.py
```diff
@@ -0,0 +1,54 @@
...
+
+logger = DaisyLogger(__name__)
+
+
+class PartnersExporter:
+
+    def export_to_file(self, file_handle, stop_on_error=False, verbose=False):
+        result = True
+        try:
+            buffer = self.export_to_buffer(StringIO())
```
neoflex: I would either pass the file_handle to the export_to method or have this method returns a string

core/management/commands/export_projects.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+from django.core.management import BaseCommand
+
+from core.importer.projects_exporter import ProjectsExporter
+
+JSON_SUFFIX = '.json'
+
+
+class Command(BaseCommand):
```
neoflex: there is a lot of common code on this three files. Many use inheritance or a common function instead

core/models/data_declaration.py
```diff
@@ -187,9 +187,27 @@ def get_long_name(self):
...
+        base_dict = {
+            "title": self.title,
+            "data_types":[ dt.name for dt in list(self.data_types)],
+            "access_category": self.share_category.name if self.share_category else None,
+            "subjects_category": self.subjects_category.name if self.subjects_category else None,
+            "de_identification":  self.deidentification_method.name  if self.deidentification_method else None,
+            "consent_status":  self.consent_status.name if self.consent_status else None,
+            "has_special_subjects": self.has_special_subjects,
+            "special_subjects_description": self.special_subjects_description,
+            "embargo_date": self.embargo_date.strftime('%Y-%m-%d') if self.embargo_date else None,
```
neoflex: maybe move the data format to a constant

PR #107, opened 2019-12-09
Feature/#106 about page

pinarpink:
I added an about page, also added to navbar.
about.html and login.html  also contains an extra paragraph on that can be uncommented for the demo deployments.

Can you please check if it is OK? Thanks.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ except ImportError:
     from distutils.core import setup
 
 requirements = [
-    'Django==2.1.11',
+    'Django==2.2',
     'django-auth-ldap==1.7.0',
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -7,7 +7,7 @@
     from distutils.core import setup
 
 requirements = [
-    'Django==2.1.11',
+    'Django==2.2',
```
neoflex: is it on purpose? it doesn't cause any incompatibility with the django extensions we are using?

neoflex:
Instead of having to uncomment part of the code, ideally I would add a flag in the settings, DEMO_MODE for instance.

PR #119, opened 2020-01-16
fix(layout): preventing tables from overflowing in datasets

jLebioda:
Could you verify before merge, please?


vildead:
All good here :) 
Tested on c00762cc06cba7e4b114957a6e6446e142e5072f

Issue #103, opened 2019-10-28
Add Project info to the string representation of Contracts


pinarpink:
**Describe the bug (current behaviour)**
Currently the string representation of contract records is the prefix. "Contract", followed by the names of signatory partners of the contract. 

**To reproduce**

1. go to https://daisy.lcsb.uni.lu
2.  go to contracts search page and use the search string "University of Tübingen"
3.  nearly a dozen contracts will be listed, all with the same string representation in their details page. This creates the problem of not being able to distinguish among contracts (with a certain institute) in contract select boxes throughout DAISY forms.

**Expected behavior**
Adding project information to the string represengtation should solve the problem.

**Environment:**

 - Deployment: daisy.lcsb.uni.lu
 - Version  1.1.0
 
 
**Additional information (error logs, screenshots)**

PR #123, opened 2020-01-23
#103 add project info

jLebioda:
Inside contracts list, project name will be appended: _Partner's name (on Project's name)_, e.g.:
Max Rubner-Institut (on Project-name)

### DIFF
--- a/core/models/contract.py
+++ b/core/models/contract.py
@@ -110,4 +110,4 @@ class Contract(CoreModel):
         else:
             project_name = "Undefined project"
 
-        return f'Contract with {partners_list} on {project_name}'
+        return f'Contract with {partners_list} - "{project_name}"'
--- a/core/models/contract.py
+++ b/core/models/contract.py
@@ -105,4 +105,9 @@ class Contract(CoreModel):
 
     def short_name(self):
         partners_list = ", ".join([p.name for p in self.partners.all()]) or "Undefined partner(s)"
-        return f'Contract with {partners_list}'
+        if self.project:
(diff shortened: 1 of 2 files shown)


vildead:
@jLebioda, could be please the project information added also to dropdown list in Data declaration editor page? That's where the user is usually struggling with long list of contracts.

@pinarpink Are there any cases/projects with more than one contract with the same collaborator? This rare case would again result in ambiguity. Contract ID could be also added to be sure...

jLebioda:
@vildead sure, will do!

jLebioda:
@vildead Could you please take a look one more time? The project information should already appear under the form.

Regarding contract-ID, could you give an example? Would it be 

    #123 Max Rubner-Institut (on Lorem Ipsum)

pinarpink:
Addition of Project info to string rep of Contract is  verified on 279f74606663a094a983d6547fc0fec279268d54.  in dev env

 @jLebioda   thank you so much. I am no expert on usability and UX,  before we merge can we replace "on" with a symbol "-"; or perhaps put the project in parentheses. Otherwise it reads like one long name.

@jLebioda @vildead  The Contract ID will be a number/label given to the contract by the relevant ServiceNOW module. Indeed there can be multiple contracts with same partner, and ID is helpful there. But, I suggest that for now we close this issue, it is best if the legal team define requirements re: contract IDs/Labels

jLebioda:
Ok, sure :)

jLebioda:
@pinarpink I've corrected it, now it reads like the following:
image

Issue #102, opened 2019-10-09
Add all model entities to Django admin


pinarpink:
**Is your feature request related to a problem? Please describe.**

We use Django admin to assist reporting over DAISY. Not all model entities have been added to the admin, e.g. Storage Locations

**Describe the solution requested**

Add all model entities to django admin.

**Describe possible alternatives**

Have a reporting  feature in DAISY
https://github.com/elixir-luxembourg/daisy/issues/99

**Other info context**

jLebioda:
Addressed in https://github.com/elixir-luxembourg/daisy/tree/102-add-all-models-to-admin

PR #124, opened 2020-01-30
chore: registered all current models in admin

jLebioda:
Addresses #102. What has been done?

- cleaned up admin.py file
- added missing models to admin.py
- added header "DAISY administration", so that it's not "Django admin" anymore
- implemented `Contract.str()` method


pinarpink:
Verified on 9361c17b1c34722fa6afb7ab384b174b0eefe7f6 in dev env.

PR #125, opened 2020-01-30
feat: make document's notes an optional field

### LINKED ISSUE #96: Make NOTE field optional when uploading document
When document notes is left empty the upload gives an error. Maybewe should make document notes optional if a document type is selected.
Note field

jLebioda:
fixes #96 

What has been done?

- changed the Document model's field to optional (it is Null=True Blank=True, and it was Null=True, Blank=False, which meant that it could be empty in database, but the application's validation will mark it as required in forms) 
- added the migration (`0010_...`)

Will require migrating the database on production and other VM's @neoflex @yjarosz


pinarpink:
Verified on  b8bcaf8dc0c1a311ca9f98af82bfc18b4191983c

PR #126, opened 2020-01-31
feat: added version number to html

jLebioda:
What has been done?

- added place for context processors (web/views/context_processors)
- refactored getting version number from view (about) to context processor; now it is available in all the templates under `{{ app_version }}`
- meta tag with version inside every page
- badge with version number on index (dashboard) page


pinarpink:
Verified on 05209fcf367a11a5955f7d1b57d265be476e6776

PR #170, opened 2020-05-11
150 new dashboard

jLebioda:
Adding the current changes to the develop


jLebioda:
Ahh, you're right @pinarpink , sorry!
I've just added it

Issue #179, opened 2020-06-04
move SECRET_KEY to settings_local.py


vildead:
@jLebioda  luckily noticed there is a security threat - SECRET_KEY variable was not changed during deployment. 

- [x] move SECRET_KEY variable to settings_local.py

- [x]  update DEPLOYMENT.md to cover change of secret key value

vildead:
Here are more things to add to settings_local.py
- [ ] import of LDAP backend modules in missing 
```
import ldap
from django_auth_ldap.config import LDAPSearch)
```
- [ ] LOGIN_* variables are missing
- [ ] `ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_NEVER)` is missing
- [ ] HAYSTACK_CONNECTIONS, HAYSTACK_SIGNAL_PROCESSOR  are missing

Issue #151, opened 2020-04-16
Update pg package URL in deployment instructions


pinarpink:
**Describe the bug (current behaviour)**

The following command in **deployment.md** no longer works as packages have moved

`sudo yum install https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-centos10-10-2.noarch.rpm`

**Expected behavior**

In deployment instructions, provide link to [postgres package url builder](https://www.postgresql.org/download/linux/redhat/) to allow deployer to find correct url.

**Environment:**

 - Deployment: https://10.240.16.89/
 - Version [e.g. 1.2.0]
 - Browser [e.g. Firefox 68.x]
 
 
**Additional information (error logs, screenshots)**

vildead:
fixed by #182

PR #182, opened 2020-06-12
181 update deployment instructions + settings_local.template.py

### LINKED ISSUE #181: Update deployment instructions
- [x] variable AUTHENTICATION_BACKENDS needs to be explained
- [x] git clone should use HTTPS instead of SSH
- [x] advise user to change the database password
- [x] separate Solr mirror troubleshooting from assignment of ownership (chown commands)
- [x] line 460 - change directory so the command works

related to #151, #179

vildead:
Closes #181 #151 #179

I have some questions:
1. `ldap.set_option(ldap.OPT_X_TLS_REQUIRE_CERT, ldap.OPT_X_TLS_NEVER)` is not included in settings template. Should it be there? I am actually not sure what that does exactly.
2. Could you please help me with the explanation of `AUTHENTICATION_BACKENDS` variable? According to #181 it should be included in deployement instructions. Should other authentication backends be included/described.?

### DIFF
--- a/elixir_daisy/settings_local.template.py
+++ b/elixir_daisy/settings_local.template.py
@@ -24,6 +24,8 @@ COMPANY = 'LCSB'  # Used for generating some models' verbose names
 # LOGIN_USERNAME_ALTERNATIVE_SUFFIX = ''
 
 # Uncomment the following lines if LDAP authentication will be used and user definitions will be bulk imported from LDAP
+# import ldap
+# from django_auth_ldap.config import LDAPSearch)
 # AUTHENTICATION_BACKENDS = ['django_auth_ldap.backend.LDAPBackend'] + AUTHENTICATION_BACKENDS
 # AUTH_LDAP_SERVER_URI = 'ldap://XXXXXXX:389'
 # AUTH_LDAP_BIND_DN = ""
--- a/elixir_daisy/settings_local.template.py
+++ b/elixir_daisy/settings_local.template.py
@@ -8,8 +8,12 @@ AUTHENTICATION_BACKENDS = [
     'guardian.backends.ObjectPermissionBackend',
 ]
 
-# SECURITY WARNING: change the secret key used in production and keep it secret !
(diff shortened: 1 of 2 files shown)


jLebioda:
@vildead FYI:
1. I'd assume that it should be there. I think it is relevant only for local deployment that uses LDAP. This very line says that TLS (maybe even TLS and SSL) should not be required when connecting to LDAP server.
2. According to the docs, it is not really crucial to tune it - i.e. the two lines in AUTHENTICATION_BACKENDS are required for the application to run; nothing to add or remove is required from the user

elixir_daisy/settings_local.template.py
```diff
@@ -6,9 +6,24 @@
     'guardian.backends.ObjectPermissionBackend',
 ]
 
+# SECURITY WARNING: change the secret key used in production and keep it secret !
+SECRET_KEY = 'qe1lmt43v1n%66vibs0&0s9qw7i!xjs^!i#f#$t_7-r8n&=+sp'
```
jLebioda, line 10: I would be slightly more in favour of:

```
L09:  # SECURITY WARNING: change the secret key used in production and keep it secret !
L10:  SECRET_KEY = None  # Generate a secret key by e.g. django.core.management.utils.get_random_secret_key()
L11:  if SECRET_KEY is None: rase NotImplementedError('You must specify SECRET_KEY in settings_local.py')
```

jLebioda (changes requested):
@vildead shall we continue with a merge?

elixir_daisy/settings_local.template.py
jLebioda: This way:
a) if the user downloads the code, tries to run it and does nothing - he will see an error with a direct message
b) if the user changes the value, the exception will not be raised

vildead: that looks good! :) I have added the change

vildead:
@jLebioda I think we are good to go. The build is failing though

jLebioda (approved):
Looks good!
Thanks

vildead:
Closes #151 
Closes #179

PR #190, opened 2020-07-23
188 discrepancies betwee import and export formats

vildead:
The exported/imported data should still be validated by [JSON schemas](https://git-r3lab.uni.lu/pinar.alper/metadata-tools/-/tree/master/metadata_tools%2Fresources) (at least by tests).

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,21 +93,19 @@ You are encouraged to try Daisy for yourself using our [DEMO deployment](https:/
 
 #### Importing 
 
-In addition to loading of initial data, DAISY database can be populated by importing Project and Dataset records from JSON files.
- The commands for import are given below: </br>
+In addition to loading of initial data, DAISY database can be populated by importing Project, Dataset and Partners records from JSON files using commands `import_projects`, `import_datasets` and `import_partners` respectively.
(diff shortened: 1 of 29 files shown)


vildead:
Features:
- import commands are unified and accept both **-d** and **-f** arguments
- partner importer added 
- JSON schema validation on import for partners, datasets and projects

There are still some issues
- [ ] There is one failing test for Submisison import :x:  I suggest to skip it for now - it will require more work - test dataset is not following schema, schema is probably not valid, importer is outdated, ...
- [ ] JSON schemas are now referenced in very clumsy way :grin: (URL to master on schema repo). I can think of submodule or pointing to tagged release (schemas should be versioned anyway).
- [ ] Readme can be updated to cover export/import commands
- [ ] some error messages could use an update
- [ ] tests of import/export commands are missing. Also more special cases of test data may be added

jLebioda:
@vildead I took your changes and pushed to a separate branch here https://github.com/elixir-luxembourg/daisy/tree/188-discrepancies-betwee-import-and-export-formats

vildead:
@jLebioda thank you! I will close this one then. The work can continue on local branch :+1:

PR #201, opened 2021-03-18
update solr schema files for docker

neoflex:
the solr schema files in the docker/solr folder were not inline with the code anymore.
I've ran build_solr_schema command to update the files


jLebioda (approved):
Looks good, let's wait for the CI to merge.

For future - to reproduce the fix we'd only need to rebuild the schema locally (`build_solr_schema`) and copy the resulting xml file into docker/solr directory?

Or would it also work if it was run inside docker and then we pulled the xml out from the container manually?

PR #216, opened 2021-05-03
Fixes to importer

vildead:
Test on DISHes on Daisy-stable
resolve conflict


vildead:
the importer was tested on daisy-stable (see logs in `owncloud/Datastewardship_R3/DAISY_Curation/DISH-copies-curate/import_logs/`)

Many warnings about missing partner - expected

pinarpink (approved):

PR #218, opened 2021-05-14
removed company_roles from contract entity and added a comment field …

pinarpink:
…to partner role


vildead (approved):
Looks good. I have tested it locally. 

Users/Datastewards will always need to add also LCSB as a signatory which is ok. I guess adding the COMPANY partner as a signatory by default is not needed, right?

Issue #230, opened 2021-05-31
Use meaningful enumeration labels for "Use Restriction Rule"


pinarpink:
Currently we're using the following enums. We should consider renaming
```
CONSTRAINTS -- > CONDITIONAL PERMISSION
NO_CONSTRAINTS --> PERMISSION
FORBIDDEN --> PROHIBITION
```

and also add `OBLIGATION`

PR #245, opened 2021-06-22
renamed use restriction enums.

pinarpink:
Fixes https://github.com/elixir-luxembourg/daisy/issues/230


core/importer/datasets_importer.py
```diff
@@ -15,7 +15,7 @@ class DatasetsImporter(BaseImporter):
     """
 
     json_schema_validator = DatasetJSONSchemaValidator()
-    json_schema_uri = 'https://raw.githubusercontent.com/elixir-luxembourg/json-schemas/master/schemas/elu-dataset.json'
+    #json_schema_uri = 'https://raw.githubusercontent.com/elixir-luxembourg/json-schemas/master/schemas/elu-dataset.json'
```
vildead, line 18: @jLebioda what do you think? Can I include the [repository with JSON schemas](https://github.com/elixir-luxembourg/json-schemas) as a submodule? It makes more sense than having one url which might not work and another fallback copy of schemas in fixtures.

vildead (approved):
Nice! The I have noticed the MR in JSON schemas is also ready ;)
Just minor: There are two migration files. Could we squish them into one?

pinarpink:
> Nice! The I have noticed the MR in JSON schemas is also ready ;)
> Just minor: There are two migration files. Could we squish them into one?

@vildead I merged migrations added migration of existing data also added the exposure of use castegory labels to JSON (fix for #250)

PR #243, opened 2021-06-16
end2end - #226 rems integration and others

jLebioda:

PR #247, opened 2021-06-24
244 add publish parameter to importer

vildead:


vildead:
- [ ] all records (datasets, partners, projects) should be first matched by elu_accession (if exists) ifelse by name ifelse by acronym
- [ ] can importer update published entity? (It wont change the elu_accesion string but can change everything else)
- [ ] check logics of entity matching:
- what if json contains elu_accession and entity does not?
- what if json does not contain elu_accesssion and entity does?
- what if json and entity have their elu_accession but they do not match? (only name or acronym matches)

pinarpink:
@vildead looking at this now sorry for the delay.

pinarpink (changes requested):
Hi Vilem I tried to verify this locally but the publishing behavior does not seem to work for me. I have put the steps I have done in the conversation tab.

pinarpink:
@vildead  

For the review: I pulled this branch. Dropped and  recreated DAISY database then  ran load_initial_data  followed by the below commands
`python manage.py import_projects --publish-entities -f  ./data/demo/projects.json`
`python manage.py import_datasets --publish-entities -f  ./data/demo/datasets.json`

Then I went to check the demo data entities from the GUI none of them had an ID assigned or appeared published.

vildead:
@pinarpink thank you for testing this. At the moment, the MR is work in progress. I took a look and you should get warning something like:
```
WARNING Publishing this type of entity (Dataset) is not implemented - item is not published. 
```
Which is actually true - the publishing feature is extended in MR [243](https://github.com/elixir-luxembourg/daisy/pull/243) which is not merged. Frankly I would like to wait till it is merged to develop rather then merging it now here.

If we need this feature soon, I can do "simple" publish and once the #243 is merged, I can update the importer to use the new function.

pinarpink:
@vildead yes this is needed before 15 July for Smart4Health

vildead:
matching and update logic:
  -  matched by external_id 
      -  db entity is published -> fail
      - db entity is not published -> continue and update all fields (and publish depending on --publish--entities flag)
  -  matched by name
      -  both have external_id
         -  db entity is published -> fail
         - db is not published -> continue and update all fields including external_id
      - only json has external_id -> update (and publish depending on --publish--entities flag)
      - only db has external_id
          - db entitye is published -> fail
          - db entitye is not published -> continue and update all fields

pinarpink (approved):
I will re-test this branch and #243 are both merged to develop

PR #249, opened 2021-06-25
limit travis build on branches develop and master

neoflex:
the goal is to limit the number of travis builds.
to be checked if it still build correctly pull requests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,4 @@ script:
 branches:
   only:
   - master
-  - stable
+  - develop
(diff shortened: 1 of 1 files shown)


jLebioda (changes requested):
Looks good - from the documentation it appears that the PRs should be triggered, too

BTW, is `daisy.fc` file relevant?

neoflex:
@jLebioda, I removed the daisy.fc file, thanks for spotting it.

jLebioda (approved):

Issue #217, opened 2021-05-14
DPIAs should be added as an attachment rather than a fixed (URL) attribute


pinarpink:
**Describe the bug (current behaviour)**

Currently the Project entity has a DPIA link attribute. This is redundant with the document upload features.
TODO:
Remove fixed DPIA attribute from Project
Add DPIA as a document type and from now on use the document upload feature to record projects' DPIAs

PR #251, opened 2021-07-02
Add DPIAs  as document attachment rather  than a fixed (URL) attribute

barrydjenaba:
Fixing issue  #217


core/fixtures/document-types.json
```diff
@@ -1,27 +1,33 @@
...
-  },
-  {
-    "name": "Subject Information Sheet Template"
-  }
+    {
+        "name": "Other"
+    },
+
+    {
+        "name": "DPIA"
```
pinarpink, line 7: Dear @nene can we replace the short hand with the expanded description i.e.  "Data Protection Impact Asssessment"

core/models/document.py
```diff
@@ -30,6 +30,7 @@ class Document(CoreModel):
                    ("consent_form", "Consent Form"),
                    ("subject_informationsheet", "Subject InformationSheet"),
                    ("project_proposal", "Project Proposal"),
+                   ('dpia', 'DPIA'),
```
pinarpink, line 33: Dear @nene can we replace the short hand  i.e. "DPIA" with the expanded description i.e.  "Data Protection Impact Asssessment"

core/models/project.py
```diff
@@ -61,9 +61,7 @@ class AppMeta:
                                    null=True,
                                    help_text='Lay summary should provide a brief overview of project goals and approach. Lay summary may be displayed publicly if the project\'s data gets published in the data catalog')
 
-    dpia = models.TextField(verbose_name='DPIA Link',
```
pinarpink, line 64: While this field is removed from the model, the migration script does not contain any corresponding change to the database. Are you sure you ran makemigrations after removing this field?

core/fixtures/document-types.json
barrydjenaba: sure, already replaced

core/models/project.py
barrydjenaba: Fixed it. thank you

pinarpink (approved):
@barrydjenaba this looks OK now. The only thing is PR #243 also has a migration script numbered 020. We might want to sort that out. Either we rebase this on develop after the merge of 243 or we merge and fix the migration script names on develop. @vildead what is your suggestion?

PR #252, opened 2021-07-07
233 export to xlsx fails for projects and datasets

vildead:


vildead:
- [ ] web/vies/export.py _do_export should be split into two functions - one getting objects specified in parameters and one creating the Excel file
- [ ] write test - fill the database with data, call every export function and see if it produces indeed valid xlsx file

vildead:
@jLebioda could you please take a look at this one too? 
 I had to update tests a bit: create new fixture for admin (user with is_staff=True) and update cohort factory
 
 I am having problem locally with test client not logging in (e.g. [here](https://github.com/elixir-luxembourg/daisy/blob/d3b448f396d3beb38972979550ceac0326d69e06/web/tests/conftest.py#L15)). The password returned by `user_data_steward.password` is a hash :/ I will be grateful for any tip. Recent test pipelines on Travis passed so I guess that's just my setup. 


 Since the login does not work, I have hardcoded the password [here](https://github.com/elixir-luxembourg/daisy/blob/d3b448f396d3beb38972979550ceac0326d69e06/web/tests/conftest.py#L20). Once the login works, that is probably not needed and `user_admin.password` should should be used. Could you please test that?
 
Big thanks!

jLebioda:
Maybe using force_login could help https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.Client.force_login ?

vildead:
Thx @jLebioda, I will not update anything in the code since this is issue with my setup :/ I have changed that to standard password retrieval.

jLebioda (approved):
Looks good!

PR #259, opened 2021-07-20
254 tracking of importer json schema versions

### LINKED ISSUE #254: Tracking of importer JSON schema versions
At the moment, JSON schemas are included locally and if not found by importer, the remote version (master) is pulled.

To be able to track which version of JSON schema is used, the tag on schema repo should be used.

@jLebioda suggested:
```
The way I see it right now is that it could be done with:
  -.  SCHEMAS_REPOSITORY = 'https://github.com/elixir-luxembourg/json-schemas/tree/${tag}/schemas/' field in the settings and SCHAMES_VERSION = '0.1.0' field in the settings
-  SCHEMAS_URI = 'https://github.com/elixir-luxembourg/json-schemas/tree/0.1.0/schemas/' field in the settings
In validator - trying to get the local file with schema, and if it fails - trying to download it using HTTPS and the address from the settings

No submodules needed, no naming changes in schema repository; and not that much changes in DAISY anyways
```

Additionally:
- [x] include test the local version is in sync with remote one (pytest)
- [x] include check on every update that the local files are updated. If not, through ERROR
- [x] validating if the settings are fine when the application is bootstrapped, and not when the import is triggered

vildead:
Closes #254 

- full JSON schema is used for validation (before, just schema URI was compared)
- schemas were updated and moved to separate folder [core/fixtures/json_schemas](https://github.com/elixir-luxembourg/daisy/tree/254-tracking-of-importer-json-schema-versions/core/fixtures/json_schemas)
- version (URI) of schema is in settings.py and it is validated on every startup ([AppConfig hook](https://github.com/elixir-luxembourg/daisy/blob/913847fe648ead0231f57fcff9e83295a4ecdc63/core/apps.py#L14))


pinarpink (approved):
the importer appears to work after the addition of schema versions. Thanks @vildead

PR #268, opened 2021-07-29
207 remove uniqueness constraint on data declaration title

vildead:


pinarpink:
I tried editing the existing data declaration "ChIP-Seq, RNA-seq", just tried changing it's title. I got this error:

'DataDeclarationEditForm' object has no attribute 'dataset'

pinarpink:
Also, I tried adding 2 new data declaration sto the Epic DREM sequencing dataset (from teh demo data). I used the name "test" for both declarations. I got the error "`Data declaration with the same title already exists for the dataset. `" when adding the second declaration. @vildead

pinarpink (changes requested):
This PR branch breaks in multiple ways on my machine. Details are in the conversation tab.

pinarpink (approved):
Editor error is eliminated. And we agreed that control for DDEC uniqueness within a DSET will be kept. Thanks @vildead .

PR #269, opened 2021-07-29
142 accession number should be unique

vildead:
there will be a conflict in migrations with #268  - I will update once merged


pinarpink (changes requested):
@vildead looking at this now, sorry I have put the comments in the wrong PR.

pinarpink:
I tried adding 2 new data declaration sto the Epic DREM sequencing dataset (from teh demo data). I used the name "test" for both declarations. I got the error "`Data declaration with the same title already exists for the dataset. `" when adding the second declaration.

pinarpink:
Also, I tried editing the existing data declaration "ChIP-Seq, RNA-seq", just tried changing it's title. I got this error:

'DataDeclarationEditForm' object has no attribute 'dataset'

pinarpink (approved):
This works as expected. Thanks @vildead

Issue #257, opened 2021-07-14
Add '--no-overwrite' parameter to importer


vildead:
Importer is updating entity when it already exists in database by default. Missing fields are either skipped or set to null.

Add flag to skip update of entities if already present (e.g. --skip-update, --no-update, --skip-overwrite, --no-overwrite, ...). If the entity already exists, it should just log a message and continue with import of other entities. 

e.g. dataset exists - should it stop and end or should it go inside the JSON and import other entities (contacts, data declarations)?

vildead:
work on https://github.com/elixir-luxembourg/daisy/tree/257-add-no-overwrite-parameter-to-importer

vildead:
@barrydjenaba the flag should be `--skip-on-exist`

PR #282, opened 2021-09-07
257 add no overwrite parameter to importer

barrydjenaba:
Fixing issue #257

### DIFF
--- a/core/importer/partners_importer.py
+++ b/core/importer/partners_importer.py
@@ -28,6 +28,9 @@ class PartnersImporter(BaseImporter):
         try:
             partner = get_partner(elu_accession=partner_dict.get('external_id', None), 
                                   name=partner_dict.get('name'))
+            if not self.skip_on_exist:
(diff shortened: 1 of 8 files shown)


vildead (approved):
Looks ok :) I have made small changes to actually invert the inner value of the flag for more clarity.

core/importer/partners_importer.py
```diff
@@ -28,6 +28,9 @@ def get_partner(elu_accession, name):
         try:
             partner = get_partner(elu_accession=partner_dict.get('external_id', None), 
                                   name=partner_dict.get('name'))
+            if self.skip_on_exist:
+                self.logger.warning(f'Partner with name \""{partner_dict.get("name")}"\" already found. It will be skipped.')
```
jLebioda, line 32: would be more readable with:
`self.logger.warning(f'Partner with name "{partner_dict.get("name")}" already found. It will be skipped.')`

but I think it can be left as it is right now

jLebioda (approved):
Looks good!

PR #283, opened 2021-09-07
fix: Submitting the data declaration form won't yield an error now ov…

### LINKED ISSUE #270: Empty use restriction form is validated
**Describe the bug (current behaviour)**
 Empty use restriction in data declaration edit form is validated - it must be first deleted in order to Submit successfully.

**To reproduce**

1. Go to Data declaration edit page
2. fill in all required fields but leave the last (extra) `restriction form` empty
3. submit

this should successfully pass since the restriction form is empty - but it is failing with "This field is required" on Use category field (required restriction_class field).

**Expected behaviour**
Since it is an 'extra' form, the validation should be skipped if the data did not change (at least according to the [doc](https://docs.djangoproject.com/en/3.0/topics/forms/formsets/#formset-validation)). But it seems the form.has_changed attribute is True in the view.

**Environment:**
bd88a9e

jLebioda:
…er empty use restrictions from the formset

Fixes #270

### DIFF
--- a/core/forms/use_restriction.py
+++ b/core/forms/use_restriction.py
@@ -33,8 +33,6 @@ class UseRestrictionForm(ModelForm):
         notes = cleaned_data.get('notes')
         if not restriction_class:
             self.add_error('restriction_class', 'Please select a valid restriction class')
-        if not notes:
-            self.add_error('notes', 'Please fill "Use Restriction notes"')
         return self.cleaned_data
 
 
--- a/core/forms/use_restriction.py
+++ b/core/forms/use_restriction.py
@@ -19,7 +19,7 @@ class UseRestrictionForm(ModelForm):
         super().__init__(*args, **kwargs)
         class_choices = [(None, "-----------------------")]
         class_choices.extend([(d.code, d.name) for d in RestrictionClass.objects.all()])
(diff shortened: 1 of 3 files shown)


core/forms/use_restriction.py
```diff
@@ -31,8 +31,10 @@ def clean(self):
         cleaned_data = super().clean()
         restriction_class = cleaned_data.get('restriction_class')
         notes = cleaned_data.get('notes')
-        if not restriction_class and notes:
+        if not restriction_class:
             self.add_error('restriction_class', 'Please select a valid restriction class')
+        if not notes:
```
vildead, line 36: We shall delete "not null" constraint on `notes` - I guess it is not a requirement and also many existing data declarations already have `notes` field empty -> to avoid errors on `edit`, we would need some kind of data migration.

vildead (changes requested):
It works almost as expected! :) 
Except validation of `notes` fields, I have one minor comment - if a user deletes one **restriction use** from many and leaves there the one `extra` form, he will get the error again (it is validated).

core/forms/use_restriction.py
vildead: null constraint in clean() removed

vildead (approved):

Issue #197, opened 2021-01-06
Flower does not work in docker-compose anymore


jLebioda:
**Describe the bug (current behaviour)**

**To reproduce**

1. Run `docker-compose up`
2. Notice in the output:

```flower_1  | ImportError: cannot import name 'Command' from 'celery.bin.base' (/usr/local/lib/python3.9/site-packages/celery/bin/base.py)
daisy_flower_1 exited with code 1
```

**Expected behavior**
Flower starts normally

**Environment:**

 - Deployment: dev deployment
 - Version 1.4
 
**Additional information (error logs, screenshots)**
probably due to a recent update in flower

jLebioda:
For an easy fix, just pinpoint the _flower's_ version (`docker-compose.yaml:71` ):
```
69. # task monitoring
70. flower:
71.      image: mher/flower
```

to

```
71.    image: mher/flower:0.9.5
```

It'd be good to investigate what changed in the new version, too

jLebioda:
This has been fixed in https://github.com/elixir-luxembourg/daisy/commit/04e3a76437d56211ee356159377e7ba3a997c50a;
however we should also look into why the new version of flower was not working and fix it (probably we'd need to update Celery)

jLebioda:
Will be fixed once #321 is merged

PR #287, opened 2021-09-08
chore: update celery, django-celery-beat and django-celery-results

### LINKED ISSUE #238: Update django-celery-results
dependabot failed to update `django-celery-results`
image (1)

jLebioda:
closes #238


jLebioda:
There's still one step we should do later - upgrade celery to v5, which will require changes in both the way celery is invoked (so docker-compose, github actions, supervisord scripts and readme file will need to be updated) and in codebase.

**Right now it will require just making a migration**

jLebioda:
After merging, #197 should be re-visited, as it will probably start working again with celery 4.4

vildead (approved):
tested on stable - works :)

PR #289, opened 2021-09-09
fix: import legal basis

vildead:


vildead:
when importing legal basis:
- list of basis with matching notes is obtained
- the list is searched for the bases with matching basis codes and data types
- if there are any, the basis is assigned only if its not general legal basis (for whole dataset)
- if there is none - new legal basis is created

barrydjenaba (approved):

jLebioda (approved):
Looks good!

Issue #263, opened 2021-07-26
Problems view


jLebioda:
**Is your feature request related to a problem? Please describe.**
It might be very handy to detect and list in one place every Entity (Dataset, Data Declaration, etc.) which has missing or invalid information.

**Describe the solution requested**
A single page accessible to a User (connected to his entities), Data Stewards/Administrators (every entity) which will show all the entities that lack correct information and require some attention.

**Describe possible alternatives**

**Other info context**
It needs a discussion about declaration what is perceived as incomplete or invalid information (1. because we already have some constraints when creating a new object of specific Entity, and 2. how to denote that lack of information is in fact what the user wanted - e.g. no document linked to a data declaration because it's a specific case and there was none)

vildead:
I should come up with issue detection conditions

vildead:
Covered by  "issues" feature from [https://github.com/elixir-luxembourg/daisy/pull/290]
next discussion goals - 

- how will be the issues accessed/presented by user?
- how to denote that lack of information is in fact what the user wanted?

PR #290, opened 2021-09-15
WIP: 264 email reporting

jLebioda:


jLebioda:
These changes contain:
 - reporting HTML and txt templates
 - primitives to gather data to report, generate the report and send the report to the users
 - a function to generate and send the reports to all the users with an email address set
 - unit tests of above

The report's contents and overall structure is not finished (and the detection of the issues is missing, too), however these changes could be already incorporated as the report is already functional

jLebioda:
I used the following to test it:

```
➜  django_celery_results git:(264-email-reportings) ✗ docker-compose exec web python manage.py shell
/usr/local/lib/python3.6/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
Python 3.6.15 (default, Sep  8 2021, 02:11:58)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.3.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from core.reporting import *

In [2]: from core.models import *

In [3]: a = User.objects.all()[1]

In [4]: generate_and_send_report_to(a)
```

jLebioda:
@vildead I believe that it can be already checked on a test instance

What was implemented?
 * `/issues` page, which shows the issues found (implemented a couple of issue checks based on what you proposed in the discussion)
 * generating the report emails
 * sending the report emails
 * periodic task definition for sending emails (to activate, it'd need to be set in both periodic tasks of django admin and settings)

core/issues.py
```diff
@@ -0,0 +1,80 @@
...
+    local_url = reverse('project', args=[str(project.id)])
+    url = get_absolute_uri(local_url)
+
+    if project.datasets.count() == 0:
+        issues.append(Issue(url, '[P-1]', 'No dataset in a project', project.acronym))
+
+    if project.has_erp == False:
+        issues.append(Issue(url, '[P-2]', 'No ERP in project', project.acronym))
+
+    if project.has_erp and len(project.erp_notes) == 0:
```
vildead, line 53: we (datastewards) will have to curate these rules - at least this one is not right

vildead (commented):
I have tested the html generation. It looks really good. 
We still have to curate issue detection rules ;)

jLebioda:
@pinarpink would you be able to take a look on that in the coming days, please?

jLebioda:
To summarise what has been done in this PR:
- a flag `EMAIL_REPORTS_ENABLED` in settings/local settings which controls if the feature is turned on/off
- a flag `should_receive_email_reports` on every User object; by default it will be turned on for any local custodian
- a html and txt template containing list of projects/datasets/issues etc. to be used in email communications
- code to find issues in datasets/projects/etc. and a special view (for every user) available under `/issues`
- a special tab for admins available under `/email_reports` to preview emails that would be sent and conveniently toggle the `should_receive_email_reports` flags
- a definition of a periodic task `test_send_email` of sending the emails to the users (i.e. to start it, it must be toggled in django-admin under `PERIODIC_TASKS`)
- unit tests and some small refactors

What we're still missing:
 - some feedback on the feature
 - a decision on what should be included in the email report (currently - issues found, some statistics and list of user's projects and datasets)
 - a definite set of the rules when there's a problem with an entity (an example of such a rule is: 
 
```
 if dataset.legal_basis_definitions.count() == 0:
        issues.append(Issue(url, '[D-1]', 'Missing legal bases for a dataset', dataset.title))
```

Screenshot 2021-10-05 at 16 54 56
Screenshot 2021-10-05 at 16 55 47
Screenshot 2021-10-05 at 16 56 09

pinarpink:
@vildead @jLebioda I made some commits to the PR branch to fix the quick issues I observed with validation rules. I also have other recommendations but they would require further work. 


- do not have separate issue tables for Project Dataset and Data Declaration. Have ONE table which is a join of all three. Have a Project, Dataset, and Declaration column in that table. consquently remove the sentence "DAISY performs some automated checks.... " from the header
- "GO TO THE OBJECT"--> "View record in DAISY"
- Test drive the report with a few of people, who use DAISY. For example, Taina or Clarissa from the Kruger group, or Xinhui, Irina, Shoiab from Biocore.

vildead:
Closes #263

aleuk0:
@vildead should we close this PR?

vildead:
@aleuk0 the feature is still on the roadmap but not a priority. We can close the PR and keep the branch if it makes things cleaner

Issue #258, opened 2021-07-20
Add flag to exporter to export also unpublished records


vildead:
At the moment, exporter is exporting all records by default.

Exporter should export only published entities. A flag "--all"/"--include-unpublished" is needed in exporter to allow export of all records.

vildead:
by default, export all records and add flag --published-only

vildead:
resolved by https://github.com/elixir-luxembourg/daisy/pull/291

PR #291, opened 2021-09-20
258 add flag to export also unpublished records

barrydjenaba:
resolving #258 
 export only exports published entities unless the --include_unpublished flag is set.


jLebioda (approved):
looks good!

vildead (approved):
looks ok :) 
I have added updates on tests and json schemas. 
We might refactor exporters in future - there is way too much repetitive code.

PR #295, opened 2021-10-01
294 add arbitrary metadata

jLebioda:
For now added an importable/exportable textual field "scientific_metadata" to Project/Cohort/Dataset/Partner (in CoreTrackedModel); it defaults to `{}`.

No validation is currently in place - we could either already merge it and add it in the (close) future for the sake of unblocking the development, or I could add some validators (thinking of RegexValidator on the field - AFAIK there's no out-of-the-box JSON validators)


jLebioda:
@neoflex @vildead it should be ready to be merged if you wanted to commence the works on the Datacatalog's side.

Figuring out a good way of dealing with validation might take me some time...

Currently thinking about implementing clean() for CoreTrackedModel as it's toggled automatically, but trying to parse JSON data in that method would actually mean that the JSON will be parsed every single time the model changes

jLebioda:
@vildead @neoflex Validation and tests for it were added, if the tests do pass, it can be already merged

jLebioda:
Note from the meeting: Maybe a JSON schema validation for the metadata could be helpful as well?

vildead:
To discuss: I think we should allow independent import of a scientific metadata. Now it is included in the entity import JSON under "metadata" key.  This can get tricky for example for dataset which has some required fields. 
 So another importer + command will be required.

vildead (approved):

PR #301, opened 2021-10-19
feat: progress in keycloak admin

### LINKED ISSUE #280: Pull account information from Keycloak programatically
Keycloak <-> DAISY integration - pulling OIDC's IDs of users

jLebioda:
closes #280


jLebioda:
@vildead I believe that we could already merge this bad boy, and address any other improvements in dedicated issues

core/utils.py
```diff
@@ -22,3 +24,25 @@ def wrap(*args, **kwargs):
...
+    """
+    def __init__(self) -> None:
+        pass
+
+    def _check_idservice(self):
+        idservice_endpoint = getattr(settings, 'IDSERVICE_ENDPOINT', None)
+        # TODO: check ID service endpoint connectivity
+        # i.e.: make http GET request 
+
+        # Checks if the function can be importable
```
vildead, line 42: small typo - can be fixed once the method is finished

vildead (approved):
Nice! First time I see abstract classes in real. And the doc of variables in settings.py is also nice addition.
I have two points we may keep for further discussions:
1. as opensource project, naming packages/modules `lcsb` does not seem right. 
2. contact and user could inherit (be) same class  - contact would just not have login right

core/utils.py
jLebioda: Ahh, adding this file was a mistake, it is by no means ready

jLebioda:
Thanks!

> 1. as opensource project, naming packages/modules `lcsb` does not seem right.
You are right! Ideally they should be in a separate django application in a separate directory. For now I've tried to keep them in a separate file so that they are not interleaved.

> 2. contact and user could inherit (be) same class  - contact would just not have login right
This is a bit confusing part - everything would be much easier if it was one class or if we could use abstractions (but databases don't like that idea)

PR #308, opened 2021-11-04
fix: increase number of use restrictions in data declarations form

### LINKED ISSUE #307: Increase the number of possible use restrictions in the daisy Data Declarations.
At the moment, the maximum number of use restrictions we can register in daisy is 10. Some datasets may have more..

jLebioda:
Closes #307


vildead (approved):
good! I hope 25 will be enough :)

PR #310, opened 2021-11-09
feat: expose contract to API

vildead:


vildead:
@jLebioda could you please have a quick look? Still some things are missing and I will have to update serializer. But any comment at this stage will be very welcomed :)

web/views/api.py
```diff
@@ -147,6 +149,38 @@ def datasets(request):
...
+def is_valid_field_in_model(klass_name, field_name):
+    getattr(klass_name, field_name, False)
+
+def get_filtered_entities(request, model_name):
+    filters = {}
+    for filter_key in request.GET:
+        if not is_valid_field_in_model(model_name, filter_key): continue
+        filters[filter_key] = request.GET.get(filter_key)
+
+    return getattr(sys.modules['core.models'], model_name).objects.filter(**filters)
```
jLebioda, line 162: this is super generic :) 
I actually thought more about passing a concrete `Contract` class to the function (see line 168), but this is even better!

core/models/contract.py
```diff
@@ -136,7 +136,7 @@ def to_dict(self):
         base_dict = {
             'id': self.id,
             'comments': self.comments,
-            'project': self.project,
+            'project': self.project.id,
```
jLebioda, line 139: are we sure that we won't break any exports by this change?

web/views/api.py
```diff
@@ -15,9 +16,10 @@
 
 from stronghold.decorators import public
 
+from core.importer.contract_exporter import ContractExporter
 from core.importer.datasets_exporter import DatasetsExporter
 from core.importer.projects_exporter import ProjectsExporter
-from core.models import User, Cohort, Dataset, Partner, Project, DiseaseTerm
+from core.models import User, Cohort, Dataset, Partner, Project, DiseaseTerm, Contract
```
jLebioda, line 22: `Contract` seems not to be used anywhere

jLebioda (commented):
Looks great! I didn't test it though yet

web/views/api.py
vildead: I am not really sure we should be that generic :D just a bad feeling it makes things too loose.

core/models/contract.py
vildead: this should be ok - contract was not used in exports yet.

web/views/api.py
vildead: fixed

web/views/projects.py
```diff
@@ -281,3 +281,14 @@ def unpublish_project(request, pk):
     project.is_published = False
     project.save()
     return redirect(reverse_lazy('project', kwargs={'pk': project.id}))
+
+
+def dsw_list_projects(request):
+    #if data steward or admin -> list all the public projects
+    # if(request.user.is_admin() | request.user.is_datasteward()):
+    #     objects = Project.objects.all().filter(is_published=True)
+    objects = Project.objects.filter(local_custodians=request.user) | Project.objects.filter(company_personnel=request.user)
```
jLebioda, line 290: from the efficiency point of view, it's better to add explicitly `is_published=True` to both queries here and remove line 291, or to move the line 291 before 290

vildead:
consider creating independent package (installable via `pip`)

vildead:
@jLebioda can you please review?
then we can merge and create new issue for moving the functionality to new package

jLebioda (approved):
looks good to me!

PR #316, opened 2021-12-14
fix: bugfix in exporter - won't export nonpublished objects anymore

jLebioda:
There were two small errors:
 * the flag `include_unpublished` was mixed up with objects parameter sometimes (when objects was missing, or `include_unpublished` was not provided as positional parameter)
 * `set_objects` lacked `self` parameter

I've changed the order of `objects` and `include_unpublished` in the exporter

@barrydjenaba Could you verify, please?

### DIFF
--- a/core/importer/datasets_exporter.py
+++ b/core/importer/datasets_exporter.py
@@ -15,8 +15,8 @@ logger = DaisyLogger(__name__)
 class DatasetsExporter:
     def __init__(
             self, 
-            include_unpublished=False,
-            objects=None):
+            objects=None,
+            include_unpublished=False):
         self.include_unpublished = include_unpublished
         """
         objects would be Django object manager containing datasets to export,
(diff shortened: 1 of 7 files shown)


barrydjenaba:
@jLebioda , I tested the export for all entities (datasets, projects, partners). After defining `include_unpublished` in `core/management/commands/export_partners.py` and `core/management/commands/export_projects.py`, the export works as expected: at the moment, only published entities are exported. Exporting all entities requires the --include-unpublished flag.

barrydjenaba (approved):

jLebioda:
> @jLebioda , I tested the export for all entities (datasets, projects, partners). After defining `include_unpublished` in `core/management/commands/export_partners.py` and `core/management/commands/export_projects.py`, the export works as expected: at the moment, only published entities are exported. Exporting all entities requires the --include-unpublished flag.

Perfect, thank you! :)

PR #317, opened 2021-12-14
chore: remove staticfiles panel from DDT

jLebioda:
Removed staticfiles panel from Django Debug Toolbar. I will need to verify it myself on a VM


vildead (approved):
I believe you :)

jLebioda:
> I believe you :)

thanks! 😉

PR #319, opened 2021-12-14
318 add citation and link to the daisy publication

vildead:


jLebioda:
@Trefex does this look good?

CITATION.cff
```diff
@@ -0,0 +1,87 @@
...
+    given-names: Regina
+    orcid: 'https://orcid.org/0000-0002-6711-8375'
+  - family-names: Alper
+    given-names: Pinar
+    orcid: 'https://orcid.org/0000-0002-2224-0780'
+  - family-names: Grouès
+    given-names: Valentin
+    orcid: 'https://orcid.org/0000-0001-6501-0806'
+  - family-names: Munoz
+    given-names: Sandrine
```
laurentheirendt, line 20: Do you have the missing ORCIDs?

CITATION.cff
```diff
@@ -0,0 +1,87 @@
+cff-version: 1.2.0
+title: >-
+  DAISY: A Data Information System for accountability
+  under the General Data Protection Regulation
+message: >-
+  If you use this software, please cite the paper
```
laurentheirendt, line 6: Maybe the `message` could be written as: `If you use this software, please cite it as below.` (i.e., not mention `paper`)

Trefex: How does this sentence work? When I'm using the software, where and how do I cite it? Shouldn't it be, "Please the below citation when referencing our work in any published materials" or something?

CITATION.cff
vildead: I have contacted Sandrine - no reply so far. Otherwise, I will just proceed without her ORCID.

CITATION.cff
vildead: I like @Trefex 's suggestion more. People will hardly cite software if they just use it. I think this was in the template I have started with. 

The message is not being displayed anyway at the  moment - there is probably default GitHub message:
Screenshot from 2022-02-04 14-45-12

vildead: fixed in [83fb454](https://github.com/elixir-luxembourg/daisy/pull/319/commits/83fb4541a562df715769da0e77c3d100965b835f)

Trefex (approved):
Once we get the ORCID from Sandrine, we can amend.

Issue #276, opened 2021-09-03
403 Local custodian cannot access document in project


vildead:
Error 403 received by local custodian when accessing document in a project 

Log:
```
EBUG 2021-09-03 15:26:33,926 checker 17454 139728188856128 [AutoChecker] Checking permission "Permissions.PROTECTED" on Document: "documents/project/2021/08/27/124_HeartMed_DMP_v3_20200911.pdf (HeartMed)" for "<name of the local custodian>": False.
WARNING 2021-09-03 15:26:33,927 log 17454 139728188856128 Forbidden (Permission denied): /documents/171/download
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/home/daisy/daisy/core/permissions/checker.py", line 194, in wrapper
    raise PermissionDenied
django.core.exceptions.PermissionDenied
```

vildead:
related to #196

PR #320, opened 2021-12-15
196 fix permission issue with document upload

### LINKED ISSUE #196: Permissions bug. Document uploader unable to view their upload
**Describe the bug (current behaviour)**

A regular DAISY user (a researcher but not a PI) can create DAISY _Project_ records and attach _Documents_ to those records. Once upload is complete the user is not able to view the attachment just uploaded. 
This bug is reported by DLSM users but observable on all deployments..

**To reproduce**

1. Go to daisy.lcsb.uni.lu login with an account that has admin priviliges
2. Go to projects page
3. Search for 'ND Collection'
4. Go to ND Collection view page and click on the eye icon to see privileges. Screenshot also attached. In this setting those local custodians that are not a VIP do not have the document view privilege.

**Expected behavior**

All users in the local custodian list should have all rights on the record. Also based on experience with users we observe that fine-grained permission handling is not at all used and better be simplified. There is a separate issue for that 

**Environment:**

 - Deployment: both DLSM and LCSB deployments
 - Version DAISY v 1.4.0

 
**Additional information (error logs, screenshots)**
Screenshot 2020-12-09 at 14 12 06

jLebioda:
closes #196

 * By default local custodians will have PROTECTED permissions now
 * Added a django command to migrate the permissions (i.e. for every single dataset and project, go through the local_custodian list and add `PROTECTED` permission if it was missing)


vildead:
@pinarpink could you please approve the local custodians should be granted a `PROTECTED` right even when they are not `VIP`?

vildead:
one tricky discuss - anyone with EDIT permission is able to add someone as local custodian (including her/himself)  thus getting access to PROTECTED content. 

Only local custodians and local custodians of an owning entity should be able to edit list of local custodians.

vildead:
option1: moderated access - datasteward would have to approve new local custodian

core/management/commands/correct_permissions.py
```diff
@@ -0,0 +1,77 @@
...
+from django.conf import settings
+from django.core.management import BaseCommand
+from core import constants
+
+from core.models import User, Dataset, Project
+from core.constants import Permissions
+
+
+class Command(BaseCommand):
+    help = 'applies the changes to permissions system (local custodians will receive PRIVILLEGED permission to a resource)'
```
vildead, line 12: lets change PRIVILEGED to PROTECTED

vildead (changes requested):
Could we please include the command granting PROTECTED permission to all local custodians as migration ?

vildead:
Closes #276

vildead:
Closes #276

PR #326, opened 2022-02-10
325 stograge entity is not exported correctly

vildead:
The platform now contains correct value - name of the backend used for storage of the data

Users with access were not listed in the exported JSON. I have added both `user` and `contact`. 

There will be more work needed:

1. exported JSON contains "storages", but in fact, these are exported from Accesses. If a storage is not set for Access instance, the record is not exported.
2. user and contact entity should be imho merged in some way. - e.g `contact` has fields `first name` and `last name` but for `user`, only `full name` is available

### DIFF
--- a/core/models/contact.py
+++ b/core/models/contact.py
@@ -87,7 +87,7 @@ class Contact(CoreModel):
             })
 
         base_dict = {
-            "pk": self.id.__str__(),
+            "pk": str(self.id),
             "email": self.email,
             "oidc_id": self.oidc_id,
             "first_name": self.first_name,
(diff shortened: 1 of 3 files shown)


vildead:
@jLebioda could you please confirm? This is just a quick fix of exporter so Valentin can continue with E2E. Thanks!

jLebioda (approved):
Looks good! 
One comment though - is there any specific reason to call:

```
"pk": self.id.__str__(),

# instead of:
"pk": str(self.id),

# or 
"pk": f"{self.id}",
```

vildead:
thanks

> Looks good! One comment though - is there any specific reason to call:
> 
> ```
> "pk": self.id.__str__(),
> 
> # instead of:
> "pk": str(self.id),
> 
> # or 
> "pk": f"{self.id}",
> ```

Thanks for review @jacek

PR #332, opened 2022-02-23
feat: small improvements for managing users in django-admin

jLebioda:
Small changes to Users section in django-admin:

 - added an optional flag to have the possibility to change users' passwords in settings.py
 - added a shortcut to django admin in the navbar
 - search bar works for both email and full_name in admin's User list
 - added columns with id and oidc_id in admin's User list (which makes the life a lot easier in case there's an user with empty full_name - because in such case the row was not clickable)


vildead (approved):
I have tested the new features - works like a charm.

PR #342, opened 2022-04-28
fix: corrected edit document icon in Project page

jLebioda:

PR #347, opened 2022-04-29
feat: removed use restriction length limitation

jLebioda:


vildead (approved):
tested locally. Thanks!
I have also resolved conflict in migrations (https://github.com/elixir-luxembourg/daisy/pull/342)

Issue #351, opened 2022-06-14
AnonymousUser is not iterable


jLebioda:
**Describe the bug (current behaviour)**
On Daisy-test, if the non-logged-in user visits the index page, the following error is produced:

![Screenshot 2022-06-14 at 18 16 18](https://user-images.githubusercontent.com/32834948/173626185-359d5123-235c-4bbe-9a08-f635a8720381.png)

**To reproduce**

1. Go to daisy-test

**Expected behavior**
Instead of the error, the dashboard should be produced

**Environment:**

 - Deployment: [e.g. daisy-test.lcsb.uni.lu]
 - Version [e.g. 1.1.0]
 - Browser [e.g. Firefox 68.x]
 
 
**Additional information (error logs, screenshots)**

jLebioda:
addressed in #353

vildead:
This has to be tested on current develop. It seems after upgrade to python 3.10, the issue is not reproducible

vildead:
not reproducibile

PR #353, opened 2022-06-15
fix: dashboard won't try to load datasets and projects for AnonymousUser

jLebioda:


jLebioda:
closes #351

web/views/dashboard.py
```diff
@@ -20,15 +23,26 @@ def dashboard(request):
...
-    # Get the notifications
-    # notifications = request.user.notifications.ordered()[:5]
+    # Return lists instead of QuerySets
+    return list(last_datasets), list(last_projects)
+
 
+def dashboard(request: HttpRequest) -> HttpResponse:
+    the_user = request.user
+
+    if the_user.is_anonymous():
```
vildead, line 33: is_anonymous is not a method but attribute 

https://docs.djangoproject.com/en/4.0/ref/contrib/auth/#django.contrib.auth.models.User.is_anonymous

vildead (changes requested):
With a small fix I get no error.

 Unfortunately, the `login` page is just skipped and I can access all records without any authentication ☹️

web/views/dashboard.py
jLebioda: my bad, I got misled by VSCode's hint; thanks!

jLebioda:
> With a small fix I get no error.
> 
> Unfortunately, the `login` page is just skipped and I can access all records without any authentication ☹️

It's actually interesting what happened with the authentication and why AnonymousUser is making his way in. I'll try to see what went wrong

vildead:
I cannot reproduce the behaviour on current `develop` (https://github.com/elixir-luxembourg/daisy/commit/95a0c1d7d90d4deffaf350e418895517ea258e33`)

Maybe we dont need to merge at all?

neoflex:
Django-stronghold doesn't work with Django 3.x if DEBUG mode is on and MEDIA_URL is not set in settings (can be set to None to fix it).
See https://github.com/mgrouchy/django-stronghold/issues/82
As Django-stronghold doesn't seem to be really maintained anymore, we could replace with https://pypi.org/project/django-require-login/.

vildead:
It will be addressed when solving issue https://github.com/elixir-luxembourg/daisy/issues/378
Now closing.

PR #354, opened 2022-07-06
feat: use table view with other fields in Contact page

jLebioda:
The result looks like this:

Screenshot 2022-07-06 at 13 42 15


vildead (approved):
Much cleaner now! Thanks!

PR #361, opened 2022-07-08
Bump Django to 3.2

vildead:


jLebioda (approved):
Will need to check it against the latest PR regarding python 3.10

jLebioda:
Yeah, it turns out that in practice the changes in this branch are canceled out by the ones from our upgrade to Python3.10

Issue #362, opened 2022-07-09
Add support for python 3.7/3.8/3.10


vildead:
Some packages we are considering for new features require python >=3.7 😸 

- [x] 3.7 - already works with last update of packages in [#361](https://github.com/elixir-luxembourg/daisy/pull/361)
- [x] 3.8 - requires bump of  `psycopg2` to `2.9.3`. With this change it works locally.
- [x] 3.10 - requires bump of `ontobio` to `2.8.2`, psycopg2 to `2.9.1`. With this change it works locally

Additional tasks:
- [x] README contains badge for each version ;)

vildead:
work on this is on branch [add-support-for-python-37-and-38](https://github.com/elixir-luxembourg/daisy/tree/add-support-for-python-37-and-38

vildead:
ontobio seems like package with many dependencies. We could replace it with integration to ontology API rather than using it just for initial import of ontology terms

vildead:
current version of celery (5.1.2) does not support python 3.10. 
We will need to bump to at least 5.2  where python 3.6 is not supported anymore -> we will need to remove support for python 3.6

PR #363, opened 2022-07-11
Add support for python 3.10

jLebioda:


vildead:
Fixes #362

setup.py
```diff
@@ -55,24 +65,21 @@
...
-    author_email='valentin.groues@uni.lu',
+    author="Pinar Alper, Valentin Grouès, Yohan Jarosz, Jacek Lebioda, Kavita Rege, Vilem Ded",
+    author_email='lcsb.sysadmins@uni.lu',
     url='https://github.com/elixir-luxembourg/daisy',
     packages=find_packages(exclude=['contrib', 'docs', 'tests*']),
     package_dir={'elixir_daisy':
                      'elixir_daisy'},
     include_package_data=True,
     install_requires=requirements,
+    dependency_links=["https://github.com/jLebioda/mockldap2.git"],
```
vildead, line 76: shouldn't this be the elixir-luxembourg repo?

vildead (changes requested):
There is one warning we may get rid of: 
Screenshot from 2022-07-26 13-30-48

Also, did you try to run `makemigrations`? I get update of id fields

setup.py
jLebioda: this should be actually gone :) Just corrected it, thanks!

jLebioda:
> There is one warning we may get rid of: Screenshot from 2022-07-26 13-30-48
> 
> Also, did you try to run `makemigrations`? I get update of id fields

Added migrations, too

vildead (approved):
Great job! Works well

jLebioda:
Excellent :)

PR #367, opened 2022-07-26
fix: importer looks for right key for storage location

### LINKED ISSUE #338: Importer fails on current schema
According to schema, there is not `locations` key in the storage. Just `location`. 

https://github.com/elixir-luxembourg/daisy/blob/77f836a37dc07300ac0250d0b02fc7742230d369/core/importer/datasets_importer.py#L152

https://github.com/elixir-luxembourg/daisy/blob/77f836a37dc07300ac0250d0b02fc7742230d369/core/fixtures/json_schemas/elu-dataset.json#L47

vildead:
Fixes #338

### DIFF
--- a/core/importer/datasets_importer.py
+++ b/core/importer/datasets_importer.py
@@ -149,7 +149,7 @@ class DatasetsImporter(BaseImporter):
                 category = self.process_location_category(storage_location_dict)
 
 
-                location_delimeted = '\n'.join(storage_location_dict['locations'])
+                location_delimeted = '\n'.join(storage_location_dict['location'])
 
                 dl = DataLocation.objects.create(
                     category=category,
--- a/core/tests/data/datasets.json
+++ b/core/tests/data/datasets.json
@@ -25,10 +25,7 @@
                         "Igor Teal"
                     ],
                     "platform": "Other",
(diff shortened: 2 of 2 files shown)


core/importer/datasets_importer.py
```diff
@@ -149,7 +149,7 @@ def process_data_locations(self, dataset, dataset_dict):
                 category = self.process_location_category(storage_location_dict)
 
 
-                location_delimeted = '\n'.join(storage_location_dict['locations'])
+                location_delimeted = '\n'.join(storage_location_dict['location'])
```
jLebioda, line 152: If it's going to be a string (and not dict anymore), `'\n'.join()` might be removed

jLebioda (approved):
are the changes reflected also in the dataset schema?

core/importer/datasets_importer.py
jLebioda: but it's no biggie, as string.join(X) works for both strings and iterables

vildead: yes! I wanted to keep that - I am not sure we wont go back to list/dict at some point ;) 
Thanks!

PR #374, opened 2022-09-13
Python3.9 downgrade

vildead:


vildead:
add OS to deployment

vildead:
For now, lets release with support of Python 3.9 only. We can add also support of 3.10 once the CI tests this version (all tests are passing locally)

neoflex (approved):

PR #376, opened 2022-09-14
set debug false by default in settings.py

neoflex:
Because of the issue with stronghold and Django 3, it became critical not to use DEBUG mode in production.


Fancien (approved):

vildead:
just to be complete - do you refer to this issue #351, right?

neoflex:
> just to be complete - do you refer to this issue #351, right?

An extension of that yes, I just created #378 to clarify.

PR #377, opened 2022-09-14
set log level to info for autoreload

neoflex:
to avoid having hundreds of lines such as:

> DEBUG File.../django/db/models/sql/compiler.py first seen with mtime XXXXXX


vildead (approved):
Nice :) I have added these lines in my settings.py already long time ago 😁

PR #379, opened 2022-09-19
Activity log history view

Fancien:
Creates a view to display the logs of audited entities (at the moment, Access only)

History is only accessible to users who has the permission to PROTECTED data associated with the logged entity.

Filters can be applied regarding date, the model logged, the field modified, or the user who did the modification. 

Buttons to link towards history are available in dataset page (in dataset header, access header and accesses table) and in the user profile page.


core/models/access.py
```diff
@@ -26,129 +26,132 @@ class Access(CoreModel):
...
+                | (Q(user__isnull=True) & Q(contact__isnull=True)),
+                name="user_or_contact_only",
             )
         ]
 
     def clean(self):
         if self.user and self.contact:
-            raise ValidationError("The Access is granted either to an User, or to a Contact. If you need both, please create two separate entries.")
+            raise ValidationError(
+                "The Access is granted either to an User, or to a Contact. If you need both, please create two separate entries."
```
neoflex, line 44: "a User"

web/views/log_entry.py
```diff
@@ -0,0 +1,129 @@
...
+    #  - Make cleaner
+    def get_list_of_model_fields(self):
+        fields_per_model = LogEntry.objects.values("content_type", "changes").distinct().all()
+        tmp_dict = {}
+        for entry in fields_per_model:
+            entry_content_type = ContentType.objects.get(pk=entry["content_type"])
+            entry_model = entry_content_type.model_class()
+            entry_model_name = entry_content_type.name
+
+            if entry_model_name not in tmp_dict:
```
neoflex, line 31: maybe that's a use case for defaultdict, no?

web/views/log_entry.py
```diff
@@ -0,0 +1,129 @@
...
+        context["end_date"] = end_date.strftime(self.DATE_FORMAT)
+        context["log_actions"] = LogEntry.Action.choices
+        context["model_fields"] = self.get_list_of_model_fields()
+        return context
+
+    def check_permissions(self, request):
+        if request.user.is_superuser or request.user.is_part_of(constants.Groups.DATA_STEWARD.value) or request.user.is_part_of(constants.Groups.AUDITOR.value):
+            return None
+        else:
+            if "entity_name" in request.GET.keys() and "entity_id" in request.GET.keys():
```
neoflex, line 116: if .GET is a dict, then `in request.GET` should already check if the key exists, no need for the `.keys()`

neoflex:
image

I have this display with the footer that I don't see in production. Is it a problem introduced with this PR?

web/views/log_entry.py
Fancien: Yes, I modified

web/views/log_entry.py
Fancien: Forgot about that.
I changed, also as I test for the presence of an entry in the keys I removed the .get() in following lines.

vildead (changes requested):
- The activity log page says: `Click on the search button to reveal the options to filter the list of history logs` but there is no `search button`. We can probably leave it out :) 
- The log is unnecessarily big (Title `Access #1` is not required at all). Could we display really simple list of changes, maybe something like this?
Screenshot from 2022-09-23 10-21-52
- The filter does not work for any combination of parameters set. I get always 403 for any search being logged in as local custodian of the owning dataset (having PROTECTED permission)

vildead (approved):
Superb!

neoflex (approved):

PR #382, opened 2022-09-21
370 json response model edit forms

Fancien:
Handle the modal EditForms with Ajax in order to display the form errors directly in the modal.

Also force the field "Remarks" in AccessForm to be updated for the form to be valid.

NB: `python manage.py collectstatic` (or `docker-compose exec web python manage.py collectstatic`) may need to be called for the changes to take effect.

### DIFF
--- a/web/views/access.py
+++ b/web/views/access.py
@@ -68,7 +68,11 @@ class AccessEditView(CheckerMixin, UpdateView, AjaxViewMixin):
         return obj.dataset
 
     def form_valid(self, form):
-        """If the form is valid, save the associated model and add to the dataset"""
+        """If the form is valid, check that remark is updated then save the associated model and add to the dataset"""
+        if "access_notes" not in form.changed_data:
+            form.add_error("access_notes", "This field must be updated")
+            return super().form_invalid(form)
+
         self.object = form.save(commit=False)
         if not self.request.user.is_anonymous:
             self.object.created_by = self.request.user
--- a/web/views/access.py
+++ b/web/views/access.py
@@ -1,15 +1,19 @@
-from core.forms.access import AccessForm, AccessEditForm
-from web.views.utils import AjaxViewMixin
-from django.views.generic import CreateView
-from core.models import Access, Dataset
-from core.constants import Permissions
-from django.shortcuts import get_object_or_404, redirect, render
+from django.views.generic import CreateView, UpdateView
+from django.shortcuts import get_object_or_404
 from django.contrib import messages
 from django.urls import reverse_lazy
-from django.http import HttpResponse, JsonResponse
+from django.http import HttpResponse
+from django.views.decorators.http import require_http_methods
+
+from core.forms.access import AccessForm, AccessEditForm
+from core.models import Access, Dataset
+from core.constants import Permissions
+from core.permissions import CheckerMixin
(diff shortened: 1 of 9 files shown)


web/views/access.py
```diff
@@ -52,33 +56,41 @@ def get_success_url(self, **kwargs):
...
-
-    log.debug(submit_url=request.get_full_path())
-    return render(request, 'modal_form.html', {'form': form, 'submit_url': request.get_full_path() })
-
-@require_http_methods(["DELETE"])
+class AccessEditView(CheckerMixin, UpdateView, AjaxViewMixin):
+    model = Access
+    template_name = 'accesses/access_form.html'
+    form_class = AccessEditForm
+    permission_required = Permissions.EDIT
```
vildead, line 63: Can we please make a check that only local custodians of the object (dataset) can create/edit the Access object?

vildead: Or staff

Fancien: So, only data steward and local custodians? Or should it include admins also?

vildead: @Fancien the ADMIN right  can be kept simple and used just for granting/removing other access rights to other users. 
Lets stick with Local Custodian (as entity level role) and datasteward (as global role).

Fancien: Ok, I added tests to check permissions on objects related to datasets (if more are added in the future, we can add them to the test)
They confirm that only user that are data stewards have Permissions.EDIT on Share, LegalBasis, Documents, DataLocation or Access. Local custodians of a dataset also have these permissions for entities related to the dataset they have custody of.

I also manually tested the website, and users who do not have this permission cannot access the edit page, whether from the modal button (not displayed for them) or from the url (request is rejected)

neoflex (approved):

web/views/access.py
vildead, line 63: Is it possible to not require only EDIT permission but also membership in either `local custodians` of the owning dataset or `DataStewardGroup` global group?
The same goes for the AccessCreateView. 

Reasoning behind - we cannot ordinary users grant privilege to create/edit access objects. (local cutodian role is exactly for that - giving notion of responsibility)

vildead (changes requested):
its working as expected. Just 2 things:
1. see the comment about extra permissions (membership) needed for create/edit access objects
2. the JSON response error is still present for Create/Edit documents modal box on Project and Contract entities (I get `{"error": {"type": "Upload error", "messages": ["No document to upload."]}}`  
3. The Remarks field still contains old string on AccessEdit form not forcing user to add a new one

web/views/access.py
Fancien: TODO:
1. For Access itself, add an additional check so that only localcustodians and data-stewards can add/edit accesses
2. To solve in future issue
3. Solved

vildead: > TODO:
> 
> 1. For Access itself, add an additional check so that only localcustodians and data-stewards can add/edit accesses
> 2. To solve in future issue
> 3. Solved

Yes! :)

Fancien: This should be solved now. Accesses have their own permission checker which returns false if user is not a datasteward or a local custodian

vildead (approved):

PR #386, opened 2022-09-23
feat: dont display created_by field on AccessEdit form

vildead:


vildead:
This is not related to the code introduced by this MR. It will be solved in https://github.com/elixir-luxembourg/daisy/issues/387

Fancien (approved):
Seems good to me

PR #388, opened 2022-09-26
Correcting accesses permissions

Fancien:
Only data-stewards and dataset local custodians can now create or edit accesses for that dataset.

Previously, anyone able to see a dataset was able to create new accesses (not edit them though).


vildead (approved):
Thanks a lot! The behaviour is fixed now. 

The **Edit** and **Create** buttons are still visible even if the user has no rights. (Its linked to EDIT permission). But that's ok. We can fix it later.

PR #389, opened 2022-10-11
Draft: Fix access validation

vildead:
The line https://github.com/elixir-luxembourg/daisy/blob/60339d6f7ec2516e9a59db7edbd1b3814dc45bce/core/models/access.py#L193 would fail with error (cannot compare `date` and  `datetime`)

Once resolved internally, the validation condition could have less **or equal** operator


vildead:
This was solved in 1.7.12

The scheduler is responsible for changing the status of access upon expiry.

PR #404, opened 2023-01-20
Changing behavior for editing accesses

Fancien:
Changed the rules to edit accesses:

- Now, anyone with EDIT permission on a dataset can can modify its accesses. Previously, only data stewards and local custodians could access the form to edit accesses
- Now, only data stewards can modify active accesses. There was no specific protection for active accesses previously

### DIFF
--- a/web/views/access.py
+++ b/web/views/access.py
@@ -90,6 +90,11 @@ class AccessEditView(CheckerMixin, UpdateView, AjaxViewMixin):
     permission_target = 'dataset'
 
     def form_valid(self, form):
+        """If access status is Active, only data stewards can edit it"""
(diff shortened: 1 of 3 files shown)


web/views/access.py
```diff
@@ -89,13 +90,19 @@ class AccessEditView(CheckerMixin, UpdateView, AjaxViewMixin):
     permission_target = 'dataset'
 
     def form_valid(self, form):
+        """If access status is Active, only data stewards can edit it"""
+        original_object = self.model.objects.get(pk=self.object.id)
+        if original_object.status == StatusChoices.active and not self.request.user.is_part_of(Groups.DATA_STEWARD.value):
```
vildead, line 95: isnt it the same validation as on line 102?

vildead:
works as a charm

PR #419, opened 2023-05-12
[Draft]Adding is_published property to dataset and rems form_name to exposure model

nirmeen:


core/forms/exposure.py
```diff
@@ -1,38 +1,38 @@
 import json
+import os
 
 import requests
 from django import forms
 from django.conf import settings
 
 from core.models import Exposure, Endpoint
+from web.views.utils import get_rems_forms
```
Fancien, line 9: It seems `json`, `requests` and `settings` are not used anymore in the file

core/forms/exposure.py
```diff
@@ -62,21 +62,19 @@ def __init__(self, *args, **kwargs):
...
-        rems_forms_response = requests.get(getattr(settings, 'REMS_URL') + 'api/forms',
-                                           headers={"x-rems-api-key": getattr(settings, 'REMS_API_KEY'),
-                                                    "x-rems-user-id": getattr(settings, 'REMS_API_USER')},
-                                           verify=getattr(settings, 'REMS_VERIFY_SSL'))
-        rems_forms_dict = json.loads(rems_forms_response.text)
+        rems_forms_dict = get_rems_forms()
         form_name_ids = []
 
         for rems_forms in rems_forms_dict:
             form_name_ids.append(
```
Fancien, line 69: Same comment as above

web/admin.py
```diff
@@ -213,6 +213,6 @@ class EndpointAdmin(admin.ModelAdmin):
 
 @admin.register(Exposure)
 class ExposureAdmin(admin.ModelAdmin):
-    list_display = ('dataset', 'id', 'endpoint', 'form_id')
-    search_fields = ('id', 'dataset', 'endpoint', 'form_id')
+    list_display = ('dataset', 'id', 'endpoint', 'form_id', 'form_name')
+    search_fields = ('id', 'dataset', 'endpoint', 'form_id', 'form_name')
     ordering = ('added',)
```
Fancien, line 218: `list_filter = ['endpoint__name']` can be added here so that datastewards can get all the exposures related to one particular endpoint.

web/views/datasets.py
```diff
@@ -58,7 +59,14 @@ def get_context_data(self, **kwargs):
         context['can_edit'] = self.request.user.can_edit_dataset(self.object)
         context['can_see_protected'] = self.request.user.has_permission_on_object(f'core.{Permissions.PROTECTED.value}_dataset', self.object)
         context['company_name'] = COMPANY
+
+        exposure_form_names_none = Exposure.objects.filter(dataset=self.object, form_name=None)
+        for exposure in exposure_form_names_none:
+            rems_forms_dict = get_rems_forms(exposure.form_id)
+            exposure.form_name = rems_forms_dict["form/internal-name"]
+            exposure.save()
```
Fancien, line 67: Is this to avoid displaying a dataset with an exposure which has no name stored?

I have the feeling this would be better to just display a placeholder character in the template if the form name is None.

We should discuss this on Monday I think

Issue #420, opened 2023-05-15
Add form name to exposures list view in dataset view and admin view


nirmeen:
**Describe the bug (current behaviour)**
exposures list in dataset details view displays only form id

**Expected behavior**

Display form name and form id in exposures list dataset details view and admin view

PR #421, opened 2023-05-15
#420-added form_name to exposures_list in dataset view and admin view

nirmeen:


web/admin.py
```diff
@@ -213,6 +213,6 @@ class EndpointAdmin(admin.ModelAdmin):
 
 @admin.register(Exposure)
 class ExposureAdmin(admin.ModelAdmin):
-    list_display = ('dataset', 'id', 'endpoint', 'form_id')
-    search_fields = ('id', 'dataset', 'endpoint', 'form_id')
+    list_display = ('dataset', 'id', 'endpoint', 'form_id', 'form_name')
+    search_fields = ('id', 'dataset', 'endpoint', 'form_id', 'form_name')
```
Fancien, line 217: We should add a filter based on `endpoint__name` here. 

Is it in another PR?

web/views/exposure.py
```diff
@@ -38,6 +38,8 @@ def form_valid(self, form):
         """If the form is valid, save the associated model and add to the exposure"""
         self.object = form.save(commit=False)
         self.object.dataset = self.dataset
+        rems_forms_dict = get_rems_forms(self.object.form_id)
```
Fancien, line 41: Should be `rems_forms_list` to be consistent

web/views/exposure.py
```diff
@@ -83,6 +85,8 @@ def get_form_kwargs(self):
     def form_valid(self, form):
         """If the form is valid, save the associated model and add to the exposure"""
         self.object = form.save(commit=False)
+        rems_forms_dict = get_rems_forms(self.object.form_id)
```
Fancien, line 88: Should be `rems_forms_list` to be consistent

web/views/exposure.py
nirmeen: this is actually a dict because i call it with the form id, so the api returns a dict with details of the form with that id

web/admin.py
nirmeen: @HesamKorki added it

web/admin.py
```diff
@@ -211,8 +211,18 @@ class EndpointAdmin(admin.ModelAdmin):
...
 
+class ReadOnlyAdmin(admin.ModelAdmin):
+    list_display = ['listing']
+
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
```
Fancien, line 222: I think you need to add `has_delete_permission`. At this point I can still delete exposures from the admin page.

image

nirmeen: i left it because i didn't see a problem in deleting an exposure from admin, do you think it should be removed?

web/admin.py
```diff
@@ -211,8 +211,18 @@ class EndpointAdmin(admin.ModelAdmin):
...
 @admin.register(Exposure)
-class ExposureAdmin(admin.ModelAdmin):
+class ExposureAdmin(ReadOnlyAdmin):
     list_display = ('dataset', 'id', 'endpoint', 'form_id', 'form_name')
+    list_filter = ('endpoint__name', 'form_id', 'form_name')
```
Fancien, line 226: Not sure filtering via `form_id` will be used. It should filter the same way as `form_name` and it is less readable I think.

nirmeen: i will remove it then

web/admin.py
Fancien: You are right, it should not cause any issue. I got stuck at 'read only' and did not think further.

Fancien (approved):
All my comments were addressed. I think it can be merged.

Issue #422, opened 2023-05-15
remove is_published field and add it as a property 


nirmeen:
**Describe the bug (current behaviour)**
storing is_published in db

**Expected behavior**

base "is_published" on dataset "exposures_list" for datasets and projects and on db for the rest of the entities

PR #423, opened 2023-05-15
#422-is-published-property

nirmeen:
Refactor 'is_published' field as abstract method and define it based on exposures_list in dataset and project, while other entities retain  'is_publishes' field with getter and setter methods.


core/migrations/0033_auto_20230515_1353.py
Fancien, line 1: Careful to change migration name and dependency once #420 is merged in develop

web/views/exposure.py
```diff
@@ -41,6 +43,8 @@ def form_valid(self, form):
         self.object.save()
         messages.add_message(self.request, messages.SUCCESS, 'exposure endpoint created')
         self.dataset.generate_elu_accession()
+        # publish subentities
+        self.dataset.publish()
```
neoflex, line 47: shouldn't we just put the call to self.dataset.generate_elu_accession() directly in dataset.publish to avoid having to call two functions and make sure the generate_elu_accession is always called with publish?

nirmeen: i felt then we need to rename the method because it is somehow different logic

neoflex: I meant keeping the generate_elu_accession method, just calling it when we call publish. As we should never publish without assigning an elu accession id, for me, it makes sense, no?

nirmeen: yes, will make it as before then

core/migrations/0033_auto_20230515_1353.py
nirmeen: done

web/views/exposure.py
nirmeen: done

Issue #432, opened 2023-07-24
Access object status does not change upon expiry


vildead:
`Active` access object should change to terminated/suspended upon reaching the expiry date. But its not the case. 

The access is not exposed to API thanks to https://github.com/elixir-luxembourg/daisy/blob/60339d6f7ec2516e9a59db7edbd1b3814dc45bce/core/models/access.py#L189C13-L189C13

but its still an issue.

+ The exposure of the dataset access should be imho dependent on the status, not on the date of expiry.

vildead:
related to https://github.com/elixir-luxembourg/daisy/pull/389

neoflex:
- add a remark, "automatically terminated" when status is changed by the scheduled task checking for expired accesses
Also with less priority:
 - remove the status suspended (migrated existing records to terminated)

neoflex:
when deploying the next version of daisy including this change, the settings_local.py file must be changed to trigger the schedulled task, see the template file as reference.

vildead:
- [ ] the fact settings_local.py needs to be updated should be included in release migration steps in DEPLOYEMENT.md file

PR #435, opened 2023-07-28
432 automate access expiration

Fancien:
Solves issue #432 by running a method every day at midnight which terminates the Access objects which have an expiration date lower than the current day.

settings_local.template.py has been modified to automatically create the Periodic Task.

### DIFF
--- a/core/models/access.py
+++ b/core/models/access.py
@@ -3,7 +3,7 @@ from typing import List
 
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Q, ObjectDoesNotExist, Count
+from django.db.models import Q, ObjectDoesNotExist, Count, signals
 
 from enumchoicefield import EnumChoiceField, ChoiceEnum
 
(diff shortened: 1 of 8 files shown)


core/tasks.py
```diff
@@ -0,0 +1,9 @@
+from datetime import date
+from celery import shared_task
+from core.models.access import Access
+
+
+@shared_task
+def check_accesses_expiration():
```
neoflex, line 7: would be good to add docstring here

core/models/access.py
```diff
@@ -38,6 +38,33 @@ class Meta:
...
+        to terminated
+
+        @param upper_date: Date threshold.
+        """
+        accesses_to_expire = cls.objects.filter(
+            grant_expires_on__lt=upper_date
+        ).exclude(
+            status=StatusChoices.terminated,
+        ).all()
+        for access in accesses_to_expire:
```
neoflex, line 54: would be good to add some logging messages

core/tasks.py
Fancien: Done.

core/models/access.py
Fancien: Done.
I added one debug log in the loop to mention accesses that are expired and to show the comparison between the expiration date and the given parameter.

I added another one at the end to say that the change was done successfully.

neoflex (approved):

Issue #431, opened 2023-07-24
Upon access approval in REMS, DAISY creates 4 access objects for given user


vildead:
This might be issue related to our current setup. But the API receiving request from REMS should be inspected.

marikapop:
@vildead can you add a DAISY screenshot with multiple access objects? Or remind me which dataset it was (I checked PRECISESADS, it doe snot have multiple objects).

marikapop:
It happened for Josie Wolfram for [PRECISESADS dataset](https://daisy.lcsb.uni.lu/dataset/169/)

neoflex:
so, I was able to reproduce the issue and it seems that the following happens:

- when daisy receives the call from REMS, it triggers a re-synchronization of users accounts from keycloak
- this takes too much times and as a result the request times out (but execution actually continues). Timeout is hardcoded to 2.5s in REMS for this request
- REMS considers then the call as failed and will retry it, which will result in multiple access created in daisy.

I propose to implement the following:

1. daisy, when receiving the call from rems, should first check if an access with the exact same information exists, if it does, just return a 200 so that rems sees it as a success
1.  check if we can safely avoid a doing synchronisation when the call is received in daisy and instead set up a scheduled task to run the synchronisation

neoflex:
When checking if an access already exists, should we take into account only accesses that were automatically granted?
For instance, if there is already an access, in active state, manually granted, and we receive a request from REMS to grant access to the same user and same dataset, what should we do? create a new access independently of the previous one, update the previously manually created access?

Another question that rose during implementation, what fields should we update exactly if identified that an access already exist and we don't want to create a new one (notes, granted_on, created_by, grant_expires_on)?

marikapop:
If we found that access for the same user to the same dataset has been manually created, I would propose to 
1. terminate this manual access by <this> date
2. add note "terminated by REMS due to a new access request by <this> date"
This will give us a single valid access object per person at any time point and keep all information about previous accesses.

Do you grant access base on status of access object or expiration date? This will influence the necessity to change expiration date of manually created access.

Valentin, could you please specify your last question? Are you asking about the same situation: REMS found manual access already exist? Or the situation when REMS finds existing REMS records?

marikapop:
@neoflex I cannot assign people. I hope you will get a notification due to @

neoflex:
here is a summary of what we discussed with @marikapop. When a new request from REMS arrives:
[...]

PR #440, opened 2023-08-02
prevent duplicated access requests creation from rems #431

neoflex:
When receiving a request to /api/rems we now check and ignore the items received if we already have active accesses created automatically for the same user, same dataset and same expiration date.

### DIFF
--- a/core/lcsb/rems.py
+++ b/core/lcsb/rems.py
@@ -36,7 +36,10 @@ else:
     synchronizer = DummyAccountSynchronizer()
 
 
-def check_existence_automatic(item):
+def check_existence_automatic(item: Dict[str, str]) -> bool:
+    """
+    Checks if an active access automatically created for the same received data already exists
+    """
     application, resource, user_oidc_id, email, expiration_date = extract_rems_data(item)
     filter_oidc = Q(user__oidc_id=user_oidc_id) | Q(contact__oidc_id=user_oidc_id)
     filter_resource_and_status = Q(dataset__elu_accession=resource, grant_expires_on=expiration_date, status=StatusChoices.active)
@@ -89,6 +92,9 @@ def handle_rems_callback(request: HttpRequest) -> bool:
 
 
 def extract_rems_data(data: Dict[str, str]) -> Tuple[str, str, str, str, date]:
+    """
+    Extracts the data from the REMS request
+    """
     application = data.get('application')
     resource = data.get('resource')
     user_oidc_id = data.get('user')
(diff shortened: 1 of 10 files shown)


core/lcsb/rems.py
```diff
@@ -155,37 +189,19 @@ def handle_rems_entitlement(data: Dict[str, str]) -> bool:
     return True
 
 
-def create_rems_entitlement(obj: Union[Access, User], 
-        application: str, 
-        dataset_id: str, 
-        user_oidc_id: str, 
-        email: str,
-        expiration_date: date) -> bool:
+def create_rems_entitlement(obj: Union[Access, User],
```
Fancien, line 192: This looks like a black-linted line. Should it not be all parameters on a new line?

core/lcsb/rems.py
```diff
@@ -81,22 +102,36 @@ def handle_rems_entitlement(data: Dict[str, str]) -> bool:
...
-        expiration_date = date.today() + timedelta(
-            days=getattr(settings, 'ACCESS_DEFAULT_EXPIRATION_DAYS', 90)
-        )
+        expiration_date = build_default_expiration_date()
     else:
         expiration_date = isoparse(expiration_date).date()
 
-    logger.debug(f'REMS :: * data access request id: {application}, user_oidc_id: {user_oidc_id}, user_email: {email}, resource: {resource}')
+    logger.debug(
+        f'REMS :: * data access request id: {application}, user_oidc_id: {user_oidc_id}, user_email: {email}, resource: {resource}')
```
Fancien, line 110: This looks like a black-linted line. Ending parenthesis should be on a new line I think

core/lcsb/rems.py
```diff
@@ -212,6 +228,26 @@ def create_rems_entitlement(obj: Union[Access, User],
     else:
         klass = obj.__class__.__name__
         raise TypeError(f'Wrong type of the object - should be Contact or User, is: {klass} instead')
-        
+
     new_logbook_entry.save()
     return True
+
+
+def build_access_notes_and_rems_user(application):
```
Fancien, line 236: Could you add some type hints to this method?

Fancien (changes requested):
Looks good. 

Feel free to ignore the black-related comments, we will probably solve all of them at the end of the sprint as you mentioned some days ago.

core/lcsb/rems.py
neoflex: I've reformatted the whole file with latest black

core/lcsb/rems.py
neoflex: I've took advantage for that to split into two functions. Didn't really made sense to have the note and the user together.

Fancien (approved):
All good now. Thank you for the changes

PR #441, opened 2023-08-03
Implement DatasetWizardView with skippable steps and refactor form handling

moustaphacheikh:
This commit introduces a multi-step form wizard view, DatasetWizardView, and a mixin, SkipFieldValidationMixin, for handling optional form steps in the wizard. The changes provide a more user-friendly and flexible way of managing the creation of datasets with data declaration, storage location, legal basis, and access 

Key changes include:

1. Implementation of **DatasetWizardView**:
    - A form_list is introduced to manage the sequence of forms in the wizard view.
    - The render_done method is refactored to handle redirection based on the current step the presence of a dataset_id in the session storage.
    - get_form_kwargs method is added to inject a dataset instance into form initialization after the first step.
    - process_step method is overridden to handle the creation and association of model instances based on form data.
    - get_context_data method is enhanced to provide additional context for template rendering.

2. Introduction of **SkipFieldValidationMixin**:
    - This mixin adds a 'skip_wizard' field to the form and overrides the is_valid method.
    - If the 'skip_wizard' field is set to True, is_valid will return True regardless of the validation of other fields in the form.
    - This allows optional steps in the wizard, enabling users to proceed without filling out the entire form.

These changes significantly improve the user experience by breaking down complex form sets into manageable, skippable steps.

Screenshot 2023-08-03 at 11 19 44

### DIFF
--- a/core/forms/dataset.py
+++ b/core/forms/dataset.py
@@ -1,8 +1,20 @@
 from django import forms
+from django.core.exceptions import ObjectDoesNotExist
 from django.shortcuts import get_object_or_404
 from django.forms import ValidationError
 from core.models import Dataset, User, Project
 from core.models.contract import Contract
+
+
+class SkipFieldValidationMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['skip_wizard'] = forms.BooleanField(initial=False, required=False, widget=forms.HiddenInput())
+
+    def is_valid(self):
+        if self.data.get(f'{self.prefix}-skip_wizard') == 'True':
+            return True
+        return super().is_valid()
 
 
 class DatasetForm(forms.ModelForm):
@@ -13,7 +25,8 @@ class DatasetForm(forms.ModelForm):
         widgets = {
             'comments': forms.Textarea(attrs={'rows': 2, 'cols': 40}),
         }
-
+        heading = "Dataset"
+        heading_help = "Please provide a help text for the dataset form"
 
     def __init__(self, *args, **kwargs):
         dataset = None
--- a/core/forms/dataset.py
+++ b/core/forms/dataset.py
@@ -1,5 +1,4 @@
 from django import forms
-from django.core.exceptions import ObjectDoesNotExist
 from django.shortcuts import get_object_or_404
 from django.forms import ValidationError
 from core.models import Dataset, User, Project
--- a/core/forms/dataset.py
+++ b/core/forms/dataset.py
@@ -4,13 +4,29 @@ from django.forms import ValidationError
 from core.models import Dataset, User, Project
 from core.models.contract import Contract
 
+from typing import Any
+
 
 class SkipFieldValidationMixin:
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """
+        Initialize the mixin. Adds a hidden Boolean field named `skip_wizard`
+        to the form which determines if validation should be skipped.
+
+        Args:
+            *args: Variable arguments to pass to the superclass.
+            **kwargs: Keyword arguments to pass to the superclass.
+        """
         super().__init__(*args, **kwargs)
         self.fields['skip_wizard'] = forms.BooleanField(initial=False, required=False, widget=forms.HiddenInput())
 
-    def is_valid(self):
+    def is_valid(self) -> bool:
+        """
+        Check if the form is valid. Validation is skipped if `skip_wizard` is True.
+
+        Returns:
+            bool: True if the form is valid or if `skip_wizard` is True, otherwise False.
+        """
         if self.data.get(f'{self.prefix}-skip_wizard') == 'True':
             return True
         return super().is_valid()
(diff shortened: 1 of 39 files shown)


web/templates/datasets/dataset_wizard_form.html
```diff
@@ -0,0 +1,69 @@
...
+                {% endfor %}
+            </ul>
+        </div>
+
+        <div class="col-md-12 card">
+            <div class="d-flex justify-content-between mt-3">
+                <a href="{% url 'wizard' %}?reset=true"
+                   class="btn btn-outline-secondary btn-raised float-right">
+                    <i class="material-icons" style="line-height:1.5;">refresh</i>
+                </a>
```
Fancien, line 26: The reset button does not fully reset the form, as the dataset that was created is not deleted.

web/tests/test_dataset.py
```diff
@@ -9,3 +12,104 @@ def test_get_dataset_add(client_user_normal):
...
+                'access-granted_on': [''],
+                'access-grant_expires_on': [''],
+                'access-access_notes': ['ssq'],
+                'dataset_wizard_view-current_step': ['access'],
+                'access-skip_wizard': [access_skip]
+            },
+            Access,
+        ],
+    }
+    for step_name, step_data in wizard_test_data.items():
```
Fancien, line 90: If you do not use `step_name`, you can call `wizard_test_data.values()` instead of `.items()`.

Also, can't the assertion be done in the other `for` loop before calling `post`. This would avoid looping twice

web/tests/test_dataset.py
```diff
@@ -9,3 +12,104 @@ def test_get_dataset_add(client_user_normal):
...
+        response = client_user_normal.post(reverse('wizard'), form_data)
+
+        if step_name != 'dataset':
+            skip_wizard_value = form_data[f'{step_name}-skip_wizard'][0]
+            if skip_wizard_value:
+                assert Model.objects.all().count() == 0
+            else:
+                assert Model.objects.all().count() == 1
+
+        if step_name == 'access':
```
Fancien, line 105: Is there another way to test that you are at the last step? This will break if we add another step after accesses

web/views/datasets.py
```diff
@@ -56,7 +126,8 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['is_admin'] = self.request.user.is_admin_of_dataset(self.object)
         context['can_edit'] = self.request.user.can_edit_dataset(self.object)
-        context['can_see_protected'] = self.request.user.has_permission_on_object(f'core.{Permissions.PROTECTED.value}_dataset', self.object)
+        context['can_see_protected'] = self.request.user.has_permission_on_object(
+            f'core.{Permissions.PROTECTED.value}_dataset', self.object)
```
Fancien, line 130: Usually in our projects, if the parameters of a function call are on a new line, the closing parenthesis should be on another new line, aligned with the function call.

E.g:
```python
method(
    param1, param2, param3
)
```

core/forms/dataset.py
```diff
@@ -5,6 +5,17 @@
 from core.models.contract import Contract
 
 
+class SkipFieldValidationMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['skip_wizard'] = forms.BooleanField(initial=False, required=False, widget=forms.HiddenInput())
+
+    def is_valid(self):
```
Fancien, line 13: Please add some type hints and docstring

core/forms/storage_location.py
```diff
@@ -1,19 +1,24 @@
...
+        heading = "Add a new storage location"
+        heading_help = "Please provide a title for this storage location"
 
     def __init__(self, *args, **kwargs):
         dataset = kwargs.pop('dataset', None)
         super().__init__(*args, **kwargs)
         # we don't allow editing dataset
         self.fields.pop('dataset')
-        self.fields['data_declarations'].choices = [(d.id, d.title) for d in dataset.data_declarations.all()]
+        if dataset:
```
Fancien, line 20: Same question as in AccessForm, is the `if` necessary?

web/tests/test_dataset.py
```diff
@@ -9,3 +12,104 @@ def test_get_dataset_add(client_user_normal):
...
     # response = client_user_normal.get(url, )
     # assert response.status_code == 302, "should redirect to first wizard step"
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('data_declaration_skip', [True, False])
+@pytest.mark.parametrize('storage_location_skip', [True, False])
+@pytest.mark.parametrize('legal_basis_skip', [True, False])
+@pytest.mark.parametrize('access_skip', [True, False])
+def test_dataset_wizard_form(client_user_normal,
```
Fancien, line 22: Please add a docstring to this test

web/views/datasets.py
```diff
@@ -16,6 +19,73 @@
 
 FACET_FIELDS = settings.FACET_FIELDS['dataset']
 
+
+class DatasetWizardView(NamedUrlSessionWizardView):
```
Fancien, line 23: Please add docstrings and type hints to this class methods

Fancien (changes requested):
Works very well, except for the issue with the reset button.

I asked for some changes, but good job :+1:

web/views/datasets.py
Fancien: Actually, ignore this.
We will run black after the sprint anyway.

Fancien:
Small comment regarding the display of the wizard steps. I feel like the skipped steps lack a bit of contrast with the white background. I would add a darker border to increase the contrast (see below).

image

This is a matter of personal preference though, feel free to ignore if you disagree :)

Fancien (approved):

PR #450, opened 2023-08-25
Fix excel export

vildead:
Some models were not serializing properly for export


moustaphacheikh (approved):
I've tested the export button on every model, and they all work properly.

Fancien (approved):

Issue #439, opened 2023-08-01
Scientific metadate on dataset level


marikapop:
Display scientific metadata field for dataset. Not in the form of dataset creation but as separate form that can be filled by a data steward.

PR #453, opened 2023-09-05
display scientific metadata field for dataset

vildead:


vildead:
Fixes https://github.com/elixir-luxembourg/daisy/issues/439

Fancien (approved):

Issue #449, opened 2023-08-25
Create button redirects to dataset creation wizard for all models


vildead:
On current develop 98a9b200a9cb41e50293a04289a788dd5614e7da clicking on Create Cohort/Contract/Project/Contact ... button (+) redirects to `https://10.240.6.196/wizard/dataset/?reset=true`

Expected behaviour would be to see corresponding model Create form

vildead:
Fixed in #454

PR #454, opened 2023-09-05
#449 - Correcting the create entity button to redirect to relevant form

Fancien:
Reverse the change to the `+` button to create entities, allowing Projects and Contacts page to correctly go to their creation forms.

~Datasets create button goes to another template where the button link points towards the wizard page~

The `add_url` context for the search page is set to the wizard in the dataset list view. This does not need any additional template. 

Related to #449


moustaphacheikh (approved):
Perfect, thanks.

PR #460, opened 2023-09-08
display all fields on dataset create form

vildead:
dataset quick form is removed
with the wizard, people will be presented with full form

### DIFF
--- a/core/forms/dataset.py
+++ b/core/forms/dataset.py
@@ -37,7 +37,13 @@ class SkipFieldValidationMixin:
 class DatasetForm(forms.ModelForm):
     class Meta:
         model = Dataset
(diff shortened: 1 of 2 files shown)


moustaphacheikh:
@vildead, Do you want the "scientific_metadata" to be accessible to all users?

vildead:
> @vildead, Do you want the `scientific_metadata` to be accessible to all users? Is the `elu_accession` only visible in edit form?

`scientific_metadata` field is already accessible to all users (with proper permissions) for project entity, so its ok for now. It can change in future though. 

`elu_accession` is displayed just in the Edit form because it cannot be changed anyway -> no need to polute the Create form/wizard. In creation phase, the attribute is always empty.

moustaphacheikh (approved):

Issue #328, opened 2022-02-11
DUO codes are not in standard format


vildead:
The suffix `[XX]` in DUO some of the codes is a placeholder for more detailed machine actionable definition of the restriction.
E.g. geographical restriction to Europe can be recorded as GS-[EU]

While this is nice feature, implementing this dynamic behaviour would require some work. 

- [x]  clean the codes and remove the optional suffixes (they make a mess in other tools where a real value is expected instead of XX).
- [ ]  record ids of the ontology terms (e.g. DUO:0000022) - labels/codes are less stable and any matching would be simpler on the ID. Align with [official DUO mapping](https://github.com/EBISPOT/DUO/blob/master/duo.csv)

thanks Danielle for great input!

vildead:
comment from Pinar:
i was meaning to create an issue somewhat related to this. What we have in Daisy is Ga4gh consent codes described in some paper. DUO is the   Follow up to consent codes and is encoded in owl ( web ontology language)
[11:18](https://lcsbbioinfocore.slack.com/archives/C9X9JCNCC/p1644574718364429)
İn short we should migrate from consent codes to duo in DAlSY.
[11:20](https://lcsbbioinfocore.slack.com/archives/C9X9JCNCC/p1644574848096429)
İn the “code” field of the db table we could keep the identifiers of DUO classes. Or whatever would be the best practice. Jacek may comment.

vildead:
related to #462 
MR removes the XX and CC placeholders

PR #462, opened 2023-09-11
clean use restriction codes

vildead:
The additional information in brackets (CC) or [XX] is just a placeholder. 

It is supposed to be replaced by code/time etc. Currently not implemented -> keeping only codes for the use restriciton categories.

### DIFF
--- a/core/models/data_declaration.py
+++ b/core/models/data_declaration.py
@@ -293,7 +293,7 @@ class DataDeclaration(CoreModel):
             "de_identification": self.deidentification_method.name
             if self.deidentification_method
             else None,
-            "consent_status": self.consent_status.name if self.consent_status else None,
+            "consent_status": self.consent_status.name if self.consent_status != ConsentStatus.unknown else None,
             "has_special_subjects": self.has_special_subjects,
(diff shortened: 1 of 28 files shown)


vildead:
Have you considered changing all codes to ontology codes as discussed in #328?

core/models/data_declaration.py
```diff
@@ -293,7 +293,9 @@ def to_dict(self):
             "de_identification": self.deidentification_method.name
             if self.deidentification_method
             else None,
-            "consent_status": self.consent_status.name if self.consent_status else None,
+            "consent_status": self.consent_status.name
+            if self.consent_status != ConsentStatus.unknown
+            else None,
```
Fancien, line 298: This means that the exported data regarding data declarations will not contain the unknown consent status anymore. Is it the intended behavior?

vildead: Thanks @Fancien for noticing. This is correct behaviour as the json schemas do not allow "unknown" value. They never had as a matter of fact :)

Fancien (approved):

PR #464, opened 2023-09-20
merging the node dependencies in the service docker image

HesamKorki:
This will simplify the deployment of Daisy + it makes the published docker image usable as a whole

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -3,6 +3,8 @@ name: Docker Image CI
 on:
   push:
     branches: [ "develop" ]
+  pull_request:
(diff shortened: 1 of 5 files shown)


Fancien (commented):
It seems to be working, I just hope that upgrade node from 16 to 18 will not cause issues later on.

The README can be slightly updated though, there is no need to say that we need to install wget in line 61 as the dockerfile now installs it.

vildead:
I am unable to properly review this MR - not enough technical skills/knowledge to to that. Lets please rely on review from @Fancien ;)

Fancien (approved):
Thank you!

PR #472, opened 2023-10-18
Updating notification models for new notification system;

Fancien:
Updating the attributes of Notification and NotificationSettings models

NotificationSetting:
- Replaced `style` by `notification_offset`
- Added `send_email`
- Added `send_in_app`

Notification: 
- Added `sent_in_app`
- Added `sent_by_email`
- Added `dismissed`
- Added `message`


notification/migrations/0004_auto_20231025_1100.py
```diff
@@ -0,0 +1,97 @@
+# Generated by Django 3.2.20 on 2023-10-25 09:00
+
+from django.db import migrations, models, connection
+import enumchoicefield.fields
+import notification.models
+
+
+def empty_notification_table(apps, schema_editor):
+    with connection.cursor() as cursor:
+        cursor.execute("TRUNCATE TABLE notification_notification;")
```
HesamKorki, line 10: I'm not sure if this is a good idea. I was thinking what if we want to move the production to another server? doesn't this mean every time this migration gets applied all the notifications are gone?

Fancien: My understanding is that the migration will not run again if the database is already at that point. 

Django stores the current version of the database in the table `django_migrations`, so if the `notification` app is already at version `0004`, it will not run this part again. 

The goal of this is to remove all notifications stored in the db which do not follow the new system (they were never used anyway)

nirmeen (approved):

PR #479, opened 2023-10-24
Import datasets, projects and partners using the ui

moustaphacheikh:
This PR introduces the `import_data` view, enabling data imports for various models using the UI.

**Highlights:**

1. **AJAX-Driven**: Ensures a smooth user experience without full-page reloads.
2. **Access Controls**: Only users in `Groups.DATA_STEWARD` can import data, ensuring restricted access.
3. **File Handling**: Efficient processing and automatic cleanup of uploaded files. 
5. **Testing**: Parameterized tests for both GET and POST requests across different model types.
Screenshot 2023-10-24 at 12 22 29
Screenshot 2023-10-24 at 12 21 04
Screenshot 2023-10-24 at 12 20 40
Screenshot 2023-10-24 at 12 20 54

### DIFF
--- a/web/templates/importer/import_modal.html
+++ b/web/templates/importer/import_modal.html
@@ -1,0 +1,16 @@
+<div class="modal fade" id="importModal" tabindex="-1" aria-labelledby="modalLabel" aria-hidden="true">
+    <div class="modal-dialog modal-lg">
+        <div class="modal-content">
+            <div class="modal-header">
+                <h5 class="modal-title" id="modalLabel">Import {{ search_url }}</h5>
+                <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                    <span aria-hidden="true">×</span>
+                </button>
+            </div>
+            <div class="modal-body px-2">
+                <div id="importForm">
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
--- a/web/tests/views/test_importer_views.py
+++ b/web/tests/views/test_importer_views.py
@@ -1,86 +1,76 @@
 import os
-from typing import Optional
+
+from django.contrib.auth.models import Group
 from django.core.files.uploadedfile import SimpleUploadedFile
 import pytest
-from django.contrib.auth.models import Group
 from django.shortcuts import reverse
-from django.test.client import Client
+
 from core.constants import Groups
-from test.factories import DataStewardGroup, UserFactory
-from core.models.user import User
 
 
-def login_test_user(
-    test_client: Client, user: User, password: Optional[str] = "test-user"
+def upload_test_file(client, model_type, user=None):
+    """
+    Helper function to upload a test file using the provided client and return the response.
+    Optionally, it logs in the given user before uploading.
+    """
+    if user:
+        client.logout()
+        client.login(username=user.username, password=user.password)
+
+    url = reverse("import_data", args=[model_type])
+
+    try:
+        with open(
+            f"{os.getcwd()}/data/demo/{model_type}.json", "rb"
+        ) as test_file_content:
+            uploaded_file = SimpleUploadedFile(
+                f"test_file_{model_type}.json",
+                test_file_content.read(),
+                content_type="application/json",
+            )
+            response = client.post(
+                url, {"file": uploaded_file}, HTTP_X_REQUESTED_WITH="XMLHttpRequest"
+            )
+        return response
+    except FileNotFoundError:
+        pytest.fail(f"Test data file for {model_type} not found!")
+
+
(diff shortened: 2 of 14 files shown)


vildead:
[data/demo/partners.json](https://github.com/elixir-luxembourg/daisy/pull/479/files#diff-3fe8eeda263d6d4fcdabb82b055f971f0c6fea4492d16e136a1efc6de29e2a66) seems to contain project not partners

vildead:
I have no other comments except the one above :)

web/views/importer.py
```diff
@@ -0,0 +1,146 @@
...
+
+    stdout_buffer = StringIO()
+    stderr_buffer = StringIO()
+
+    log_buffer = StringIO()
+    log_handler = logging.StreamHandler(log_buffer)
+    LOGGING_FORMAT = "%(levelname)s: %(message)s"
+    log_formatter = logging.Formatter(LOGGING_FORMAT)
+    log_handler.setFormatter(log_formatter)
+    logger = logging.getLogger()
```
Fancien, line 117: This returns the default logger right? Would it not be better to get the logger associated with the command, just in case it does not use the default one?

Fancien (changes requested):

web/views/importer.py
moustaphacheikh: i'm not sure there is one logger that i can targeted here

web/tests/views/test_importer_views.py
```diff
@@ -1,86 +1,76 @@
...
-from core.constants import Groups
-from test.factories import DataStewardGroup, UserFactory
-from core.models.user import User
 
-
-def login_test_user(
-    test_client: Client, user: User, password: Optional[str] = "test-user"
-):
-    assert test_client.login(username=user.username, password=password), "Login failed"
+from core.constants import Groups
```
Fancien, line 8: Same question as above

web/tests/views/test_importer_views.py
```diff
@@ -1,86 +1,76 @@
 import os
-from typing import Optional
+
+from django.contrib.auth.models import Group
```
Fancien, line 3: This is not used in the file. Any reason why you import it?

web/views/importer.py
Fancien: Yes, the issue is that even if there is a specific loader set, there is no way to guess what the logger name will be :/

web/tests/views/test_importer_views.py
```diff
@@ -0,0 +1,86 @@
...
+@pytest.mark.parametrize(
+    "group, model_type",
+    [
+        (DataStewardGroup, "projects"),
+        (DataStewardGroup, "datasets"),
+        (DataStewardGroup, "partners"),
+    ],
+)
+@pytest.mark.django_db
+def test_importer_post_request(group, model_type):
```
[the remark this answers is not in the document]
moustaphacheikh: Can you elaborate more? Do you mean if the user submits non-json file?

web/templates/importer/import_modal.html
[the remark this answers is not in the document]
moustaphacheikh: i don't know if we use the same pattern again we can make a template for it.

web/tests/views/test_importer_views.py
moustaphacheikh: done

web/tests/views/test_importer_views.py
```diff
@@ -0,0 +1,76 @@
...
+    # Test POST request for data steward
+    response = upload_test_file(client_user_data_steward, model_type)
+    assert response.status_code == 200
+
+    # Test GET request for data steward
+    url = reverse("import_data", args=[model_type])
+    response = client_user_data_steward.get(url, HTTP_X_REQUESTED_WITH="XMLHttpRequest")
+    assert response.status_code == 200
+    assert "importer/import_form.html" in [t.name for t in response.templates]
+
```
Fancien, line 59: The file is a demo one right? Would it be possible to check that the dataset/project/partner is correctly imported in the db?

moustaphacheikh: The success field only indicates if an error occurred, not if the import was correct. This feature solely returns the output to the user. testing if the file was imported correctly in the db is done in the importers tests.

Fancien: Ok, if the test that the object is in the db is already there that's good.

Checking that the view returns `{success: True}` would still be useful. It would mean that we would expect the import of the test file to be successful, whatever 'successful' means behind the scene.

Fancien (approved):

PR #483, opened 2023-10-27
Restrict visibility of "Save the results as xlsx" and "Import" buttons

moustaphacheikh:
Restrict visibility of "Save the results as xlsx" and "Import" buttons to Data Stewards only.

### DIFF
--- a/web/templates/_includes/search_form.html
+++ b/web/templates/_includes/search_form.html
@@ -6,7 +6,7 @@
         </h1>
     </div>
     <div class="col-5 d-flex align-items-center justify-content-end">
-        {% if request.user|has_group:"daisy-data-steward" and not request.user.is_superuser%}
+        {% if request.user|has_group:"daisy-data-steward" %}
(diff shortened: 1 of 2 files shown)


nirmeen:
@moustaphacheikh should we add to the export methods that the user is authenticated and is data Stewards?

web/templates/_includes/search_form.html
```diff
@@ -6,15 +6,13 @@ <h1 class="display-4">
         </h1>
     </div>
     <div class="col-5 d-flex align-items-center justify-content-end">
-        {% if request.user|has_group:"daisy-data-steward" %}
+        {% if request.user|has_group:"daisy-data-steward" and not request.user.is_superuser%}
```
Fancien, line 9: I think this condition means Vilem cannot see the button. Superusers are not datastewards by default, and this condition is specifically removing superuser from the list of people allowed to see the import and export buttons

Fancien (approved):

PR #485, opened 2023-10-27
Email notification tasks and templates #472

nirmeen:
Implement email notification tasks:

- report_notifications_upcoming_events_errors_for_datasteward
- send_notifications_for_user_upcoming_events
- send_missed_notifications_for_user_upcoming_events

Implement email templates to send a listing of notifications by entity name


notification/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+            )
+        except Exception as e:
+            logger.error(
+                f"Failed: An error occurred while sending Email notification error report for data-stewards."
+                f" Error: {e}"
+            )
+            print(
+                f"Failed: An error occurred while sending Email notification error report for data-stewards."
+                f" Error: {e}"
+            )
```
Fancien, line 61: This could be removed no? Also, is it intentional for the error to not be raised again?

notification/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+                notif.save()
+        except Exception as e:
+            logger.error(
+                f"Failed: An error occurred while sending upcoming events Email notification for user {user.full_name}."
+                f" Error: {e}"
+            )
+            print(
+                f"Failed: An error occurred while sending upcoming events Email notification for user {user.full_name}."
+                f" Error: {e}"
+            )
```
Fancien, line 124: Same as above, is the `print` necessary?

notification/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+        finally:
+            # report error to data-stewards
+            report_notifications_upcoming_events_errors_for_datasteward(user)
+            continue
+
+
+@shared_task
+def send_all_notifications_for_user_upcoming_events(execution_date: str = None):
```
Fancien, line 132: I feel this is the exact same task as above, except for the filter applied to the Notification queryset. 

Would it make more sense to add a boolean parameter to the task (something like `one_day_only` which defaults to True for example)? When the parameter is true we set the filter `time__date=exec_date`, otherwise we set the filter `time__date__lte=exec_date`

nirmeen: yes i had the same thought. I will do so. Thank you.

notification/tasks.py
nirmeen: yes it could be removed. The idea was that we log the error and then send datastewards an email in case of notification error and continue to the next user. you think we should raise the error anyways?

Fancien: It makes sense. I think the error email should be sent at the very end though, or we risk sending an email for every user on some days.

notification/tasks.py
nirmeen: done

notification/tasks.py
nirmeen: i thought it is one email per user. I felt it would be easier for data stewards to track which user notifications errors they checked and which they didn't. Otherwise. if all users in one email it might be long and maybe they will not be able to keep track of where they stopped.

elixir_daisy/settings_local.template.py
```diff
@@ -84,5 +85,9 @@
     "clean-accesses-every-day": {
         "task": "core.tasks.check_accesses_expiration",
         "schedule": crontab(minute=0, hour=0),  # Execute task at midnight
-    }
+    },
+    "notifications-email-every-day": {
+        "task": "notification.tasks.send_all_notifications_for_user_upcoming_events",
```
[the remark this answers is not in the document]
nirmeen: yes, I will update it.

notification/templates/notification/email.html
```diff
@@ -15,26 +18,29 @@
...
+    <table role="presentation" border="0" width="100%" cellspacing="0" style="color: #212529; font-family: sans-serif; font-size: 16px;">
+        {% block content %}{% endblock %}
+    </table>
+    <!-- footer -->
+    <table role="presentation" border="0" width="100%" cellspacing="0">
+        <tr>
+            <td style="color: #212529; font-size:13px; padding-left: 10px;" width="100%">
+                    {% block footer %}
+                    This email has been generated from <a href="{{server_url}}" style="color: #f47d20;">DAISY</a>.
+                    You can change your notifications settings in your <a href="{{profile_url}}" style="color: #f47d20;">profile</a>.
```
Fancien, line 40: This can probably use `{% url 'profile' %}` instead. Note that I changed the url in the PR for the notifications views

Fancien: Ok, this uses a variable instead of {% url %} to avoid defining the url in both .txt and .html templates. So that makes sense to leave it as it is.

elixir_daisy/settings_local.template.py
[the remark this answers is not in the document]
nirmeen: done.

notification/tests/test_tasks.py
```diff
@@ -0,0 +1,157 @@
+from datetime import datetime, timedelta
+
+import pytest
+
+from notification import tasks
+from notification.models import NotificationSetting, Notification
+from test.factories import *
+from django.test import override_settings
+
+
```
Fancien, line 10: I would add one additional test to check that when `only_one_day` is True, the notifications from the previous days are not processed.

nirmeen: done

elixir_daisy/settings_local.template.py
```diff
@@ -79,10 +80,20 @@
 # KEYCLOAK_USER = 'your service user for daisy'
 # KEYCLOAK_PASS = 'the password for the service user'
 
+# Email related setting
+EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
```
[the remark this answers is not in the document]
nirmeen: removed it.

Fancien (approved):
Thanks for the changes!

notification/tasks.py
nirmeen: sorry, removed it now.

Issue #455, opened 2023-09-06
Display scientific metadata field only to datastewards


vildead:
The scientific_metadata field of Dataset and Project is rather hacky solution for now. 

As such, it shall be displayed only to datastewards.

Related to https://github.com/elixir-luxembourg/daisy/issues/439

vildead:
done

PR #486, opened 2023-10-27
#455 - Removing scientific metadata field for non data stewards

Fancien:
Dataset and Project views now set a `keep_metatada_field` in the form kwargs.

If that field is not set with a value of True, the DatasetForm and ProjectForm removes `scientific_metadata` from their fields.

### DIFF
--- a/web/views/datasets.py
+++ b/web/views/datasets.py
@@ -1,11 +1,11 @@
 from django.conf import settings
-from django.shortcuts import render, redirect, get_object_or_404
+from django.shortcuts import render, get_object_or_404
 from django.urls import reverse_lazy
 from django.contrib import messages
 from django.views.generic import CreateView, DetailView, UpdateView, DeleteView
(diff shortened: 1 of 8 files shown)


web/views/datasets.py
```diff
@@ -78,6 +78,10 @@ def get_form_kwargs(self, step: str = None) -> Dict[str, Any]:
                 Dataset, pk=self.storage.extra_data.get("dataset_id")
             )
             kwargs["dataset"] = dataset
+        elif self.request.user.is_part_of(Groups.DATA_STEWARD.value):
+            # If the user is a data steward, we want to keep the scientific metadata field
+            kwargs["keep_metadata_field"] = True
+
```
nirmeen, line 84: Should we add user_can_access_scientific_metadata method just in case the group of users changes?

nirmeen: Thank you

nirmeen (approved):

PR #488, opened 2023-11-03
refactor keycloak synchronization logic

neoflex:
## Implementation notes

- Add celery task to run full daisy keycloak synchronization
- Add celery beat configuration in settings template
- Stop triggering full synchronization when receiving request on REMS endpoint
- Re-synchronize only the user or contact corresponding to the REMS request
- When matching contact or user is found in daisy, update first name, last name and email in daisy with data coming from KC.

## Flow 

Untitled-2023-10-24-0926


core/synchronizers.py
```diff
@@ -28,16 +44,121 @@ def test_connection(self) -> None:
...
 
+    @abstractmethod
+    def get_external_user_info(self, oidc_id: str) -> Dict[str, str]:
+        """
+        Should return a dictionary with the user information
+        """
+        pass
+
+
+def update_user_or_contact(
```
Fancien, line 55: It feels strange to have methods called `update_user_or_contact` both inside and outside the AccountSynchronizer.

Is there a reason why these had to be defined outside the class?

core/lcsb/oidc.py
```diff
@@ -6,22 +6,29 @@
 from django.conf import settings
 from django.db import transaction
 from keycloak import KeycloakAdmin
+from keycloak.exceptions import KeycloakGetError
 
 from core.models.contact_type import ContactType
 from core.models.contact import Contact
 from core.models.partner import Partner
 from core.models.user import User
```
nirmeen, line 14: those imports are not used. along with 

import json
from multiprocessing import Lock
from django.db import transaction
from core.synchronizers import AccountSynchronizationException,

core/synchronizers.py
```diff
@@ -28,16 +44,121 @@ def test_connection(self) -> None:
...
+    contact.partners.add(partner)
+    contact.save()
+    return contact
+
+
+def update_and_save_entity(entity, info):
+    entity.last_name = info.get("last_name", entity.last_name)
+    entity.first_name = info.get("first_name", entity.first_name)
+    entity.email = info.get("email", entity.email)
+    entity.oidc_id = info.get("id")
```
nirmeen, line 138: The entity is not saved here, do you think it is feasible to rename the method?

core/synchronizers.py
```diff
@@ -47,23 +168,122 @@ def test_connection(self) -> bool:
...
+            )
+
+    def retrieve_and_update_user(self, external_id) -> Optional[User]:
+        user_info = self.get_user_info(external_id)
+        return update_user(user_info, external_id)
+
+    def retrieve_and_update_contact(self, external_id) -> Optional[Contact]:
+        contact_info = self.get_contact_info(external_id)
+        return update_contact(contact_info, external_id)
+
```
nirmeen, line 284: are these used/needed?

web/views/api.py
```diff
@@ -23,11 +23,8 @@
 from core.models import (
```
Fancien, line 23: At the very top of this file, the `import os` is not needed it seems

web/tests/test_rems.py
```diff
@@ -47,9 +113,25 @@ def test_rems_handler_duplicate(client, user_vip, user_data_steward):
     assert len(accesses) == 1
 
 
-def test_rems_handler_no_expiration(client, user_vip, user_data_steward):
+def test_rems_handler_no_expiration(client, user_vip, user_data_steward, mocker):
     resource_id = "TEST-2-5591E3-1"
     email = "john.doe@uni.lu"
+
+    def mock_get_external_user_info(self, oidc_id):
```
Fancien, line 120: It feels to me like this whole command could go outside the tests, to avoid repeating the definition.

The only thing that differs between the calls are the first name, last name, and email of the mock user. Could they be set as the method argument?

core/lcsb/oidc.py
```diff
@@ -86,240 +93,89 @@ def get_list_of_users(self) -> List[Dict]:
...
+                "id": user.get("id"),
                 "email": user.get("email", None),
-                "first_name": user.get("firstName", "FIRST_NAME_MISSING"),
-                "last_name": user.get("lastName", "LAST_NAME_MISSING"),
-                "username": user.get("email", "USERNAME_MISSING"),
+                "firstName": user.get("firstName"),
+                "lastName": user.get("lastName"),
             }
             for user in keycloak_response
             if user.get("emailVerified", False)
```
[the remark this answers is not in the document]
neoflex: it should yes, I kept it as a double check in case the filter above stop working for some reasons, new keycloak version for instance.

core/synchronizers.py
neoflex: thanks for spotting this, will fix

core/synchronizers.py
neoflex: no, I'll remove. thanks

core/lcsb/oidc.py
neoflex: thanks, imports removed

core/synchronizers.py
neoflex: I renamed the function and added docstring

core/synchronizers.py
neoflex: done

web/views/api.py
neoflex: removed, thanks

web/tests/test_rems.py
neoflex: done

core/synchronizers.py
nirmeen: thanks

Fancien (approved):

PR #489, opened 2023-11-06
Implement notification views

Fancien:
Adding views for notifications, notification settings and admin notifications page.

Adding api endpoints to allow browser to request list of notifications.

Notifications list and cards are built using React and React-table


web/templates/navbar.html
```diff
@@ -53,6 +53,16 @@
             </ul>
             {% if request.user.is_authenticated %}
                 <ul class="navbar-nav">
+                    <li class="nav-item {% if request.resolver_match.url_name == "notifications" %}active{% endif %}">
```
nirmeen, line 56: should we disable/remove the notifaction button when NOTIFICATIONS_DISABLED is true?

web/views/notifications.py
nirmeen, line 1: this file can be removed. seems not used.

nirmeen:
Can the README be updated?

notification/templates/notification/notification_list.html
```diff
@@ -0,0 +1,48 @@
...
+{% block content %}
+    {% csrf_token %}
+    <div class="row mt-4">
+        <div class="jumbotron col">
+            <h1 class="display-4">Notifications</h1>
+            <hr />
+            <p>If you would like to change the way DAISY handles your notifications, please update your <a href="{% url 'notifications_settings' %}">notifications settings</a>.</p>
+            {% if not show_dismissed %}
+                <p>By default, dismissed notifications are not shown. You can display dismissed notifications by clicking <a href="{{ request.path }}?show_dismissed=true">here</a></p>
+            {% endif %}
```
neoflex, line 18: would be convenient for the user to have an easy way to hide the dismissed notification.
Adding a similar text at the one above but with a link for hiding them for instance.

notification/views.py
```diff
@@ -1,3 +1,94 @@
...
+from django.views.generic import UpdateView, ListView
+from django.urls import reverse_lazy
+from django.shortcuts import get_object_or_404, reverse, redirect
+from django.core.exceptions import ObjectDoesNotExist
+from django.contrib.auth.mixins import UserPassesTestMixin
+from django import forms
+
+from notification.models import Notification, NotificationSetting
+from core.models import User
+from web.views.utils import is_data_steward
```
neoflex, line 12: import is not used it seems

neoflex:
I find it a bit confusing to have an entry called "Profile" in the menu but then the page's title is "Notification settings".
If we don't have anything else in the section, we could maybe just put "Notifications settings" in the menu.

notification/views.py
```diff
@@ -1,3 +1,97 @@
...
+        context["admin"] = True
+        return context
+
+    def post(self, request, **kwargs):
+        form = self.UserSelection(request.POST)
+        if form.is_valid():
+            user = form.cleaned_data.get("user")
+            return redirect("notifications_admin_for_user", pk=user.pk)
+        else:
+            return redirect("notifications_admin")
```
neoflex, line 97: when switching user, we looks the "show dismissed" setting. It's can be an annoyance if the admin has to reclick on the link after switching users if they do that often.

notification/templates/notification/notification_list.html
nirmeen, line 18: The links can be surround by <strong> tags or have a font weight bold similar to user guide link in homepage
Screenshot 2023-11-08 at 14 11 26

web/templates/navbar.html
Fancien: I think it would make sense yes. I will check how to change that

web/templates/navbar.html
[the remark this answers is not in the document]
Fancien: Yes, done

notification/views.py
Fancien: Yes, I removed it

notification/views.py
Fancien: This should be solved in latest commit

notification/templates/notification/notification_list.html
Fancien: Corrected

Fancien: I added that in latest commit

web/views/notifications.py
Fancien: Yes, I removed it in the latest commit

web/templates/navbar.html
Fancien: Corrected

web/static/vendor/components/notification/notification_table.tsx
```diff
@@ -0,0 +1,169 @@
...
+        }),
+        // The objectName column displays the name of the object that triggered the notification
+        columnHelper.accessor("objectName", {
+            header: () => {
+                const title = notifications[0].objectType;
+                return title.charAt(0).toUpperCase() + title.substring(1);
+            },
+            cell: cell => {
+                const text = cell.getValue();
+                const displayedText = text.length < 70 ? text : text.substring(0, 70) + "...";
```
nirmeen, line 78: would it make since to truncate it by css instead and hover to see tooltip? or if it is too much work not to truncate it and display it in full. My thought is that description message can't be found in the Entity details page so if it has important info to the user it should be there.

web/static/vendor/components/notification/notification_card.tsx
```diff
@@ -0,0 +1,79 @@
...
+ */
+export const NotificationCard = ({title, type, newNotifNumber, children, showDismissBtn, dismissAll}: NotificationCardProps) => {
+    return (
+        <div className={"row mt-4 accordion"}>
+            <div className={"card col px-0"}>
+                <NotificationHeader className={title} category={type} showDismissNumber={showDismissBtn} newNotifications={newNotifNumber} />
+                <div id={`accordion-body-${type}`} className={"collapse p-3"}>
+                    {showDismissBtn &&
+                        <div className={"d-flex justify-content-end"}>
+                            <a className={"btn btn-link btn-outline float-right"} onClick={() => dismissAll(type)}>Dismiss all</a>
```
nirmeen, line 67: Can we hide dismiss all button when notifications are already dismissed. 
To reproduce, click on dismiss all button and then click on show dismissed, dismiss all button appears in the notifications accordion.

notification/templates/notification/notification_list.html
nirmeen: thank you!

web/templates/navbar.html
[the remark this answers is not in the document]
nirmeen: thank you : )

web/views/notifications.py
nirmeen: thanks

web/templates/navbar.html
nirmeen: I am not sure if we should do the same for notifications_settings and admin_notifcations, what do you think ?

nirmeen: In the settings it is commented that 
```
# by default, notifications by email are disabled
NOTIFICATIONS_DISABLED = True
```

So it seems that that isn't for the in_app notifications.

web/static/vendor/components/notification/notification_table.tsx
Fancien: I cannot truncate the text via css without fixing the width for the columns. The issue is that I do not understand how to do that in the current react-table version (it was much easier in the version used for the datacatalog).

I can however add the hover box relatively easily I think.

web/templates/navbar.html
Fancien: True, but it would still make sense to remove the buttons if the notifications are not enabled. I will also put these menus under a `if` condition

web/static/vendor/components/notification/notification_table.tsx
nirmeen: it works, thank you!

web/static/vendor/components/notification/notification_card.tsx
nirmeen: works, thank you!

web/templates/navbar.html
nirmeen: works, thanks.

neoflex (approved):

PR #492, opened 2023-11-16
Notification views feedback implementation

Fancien:
- Avoiding an issue where having the notification app disabled would create a lot of error messages in the JS console
- Enabling the notification app by default
- Showing the number of non-dismissed notifications in the navbar even on the notifications page
- Changing the column names in the Notification tables
- Changing the `Dismiss` buttons display in the Notification tables
-  Moving the accordion collapsing system inside the NotificationCard component
- Adding an indicator in the card headers to show that the header is clickable
- Moving the card header number badge closer to the card title and changing its color to red
- Replacing the links to show/hide dismissed notifications by a switch toggler and linking react state to it 
- Keeping the `checked` status of the switch when changing the user filter (in the admin view of notifications)


Fancien:
Small issue with the `Show dismissed` switch that needs working on. The rest should be fine for review

notification/static/notification/css/notification.css
```diff
@@ -1,5 +1,22 @@
 .card-badge {
     position: absolute;
-    right: 25px;
+    right: -25px;
     bottom: 35%;
```
nirmeen, line 4: I suggest to make accordion title centred vertically by setting margin-bottom: 0; and adjusting card-badge bottom:0; to make it centred too.

nirmeen:
@Fancien All works smoothly, great job. One thing to add is, admin notifications in navbar should be accessible only to data Stewards.

notification/static/notification/css/notification.css
Fancien: I'll check, thanks!

Fancien: It should be solved now. Could you have a look?

PR #493, opened 2023-11-17
add display_name method to notifyMixin and remove days from msg

HesamKorki:


notification/__init__.py
```diff
@@ -82,3 +82,9 @@ def get_absolute_url(self) -> Optional[str]:
         Returns the absolute url of the entity.
         """
         return None
+
+    def display_name(self) -> str:
+        """
+        Returns the display name of the entity.
+        """
+        return self.get_absolute_url()
```
Fancien, line 90: Wasn't it supposed to be self.__str__() by default?

HesamKorki: You are absolutely correct. This is been addressed in the latest commit

core/models/document.py
```diff
@@ -78,6 +80,13 @@ class Meta:
         null=True,
     )
 
+    def get_absolute_url(self) -> str:
+        model = self.content_type.model
+        return reverse(model, args=[str(self.object_id)])
+
+    def display_name(self) -> str:
+        return self.content.name.split("/")[1]
```
Fancien, line 88: Would it make sense to just return the filename here? Current implementation returns the first folder in the path to the object.

```suggestion
        return self.shortname
```

Fancien (commented):
I modified the Notification `to_json` method to use the new display_name method.
Also, I added a suggestion for the Documents display_name implementation, feel free to ignore it.

core/models/access.py
```diff
@@ -250,6 +250,12 @@ def is_active(self):
 
         return True
 
+    def display_name(self) -> str:
+        return f"{self.status}|{self.dataset}|{self.user or self.contact}"
```
nirmeen, line 254: Should a space be added before and after the "|"? that is how it currently looks
Screenshot 2023-11-21 at 12 10 56

and that is how it would look with a space.
Screenshot 2023-11-21 at 12 12 42

core/models/document.py
nirmeen: should we also add the entity name for the display to be like "EPIC-DREM | test.xml"? because with the file name only it feels like a link to the file itself not to the project/contract details page

Fancien (approved):
Thanks for the changes @HesamKorki

PR #502, opened 2024-02-06
1.8.0

vildead:


vildead:
The release notes are drafted: https://github.com/elixir-luxembourg/daisy/releases

I will test deployment on stable instance and we are ready to merge

vildead:
Migration on daisy-stable was successful with no errors.

PR #507, opened 2024-05-22
Hot/fix OIDC contact

HesamKorki:
There is a new validation check for the contact form that denies users from changing the oidc of a contact


vildead:
@HesamKorki could you please resolve the conflict? I will approve once done :)

PR #511, opened 2024-05-29
Update setup.py to contain current version

vildead:


vildead:
@moustaphacheikh could you please take a look at this small change? Thanks!

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    - name: Validate version in setup.py
+      run: |
+        VERSION=$(python setup.py --version)
+        echo "Version in setup.py: $VERSION"
+        if [[ "$VERSION" != "$GITHUB_REF_NAME" ]]; then
```
moustaphacheikh, line 24: This means that the release format will change from **v1.8.1** to **1.8.1** ?

vildead: nice catch! I will add the `v` to the $VERSION variable

vildead: fixed @moustaphacheikh  :)

moustaphacheikh (approved):

PR #521, opened 2024-09-17
Add support for backup and restore using docker

moustaphacheikh:


docker-compose.yaml
```diff
@@ -84,10 +96,43 @@ services:
...
+    # Backup service
+    backup:
+        image: alpine:3.20.3
+        environment:
+            - DB_HOST=db
+            - DB_PORT=5432
+            - DB_NAME=${DB_NAME:-daisy}
+            - DB_USER=${DB_USER:-daisy}
+            - DB_PASSWORD=${DB_PASSWORD:-daisy}
+            - BACKUP_DIR=/backups
```
vildead, line 111: can we specify this in the env file? 
Can this be an external folder? (outside of the container)

vildead:
I still need to go through the README.md and DEPLOYMENT.md to avoid overlap with /doc folder.

docker-compose.yaml
moustaphacheikh: `BACKUP_DIR` refers to the directory inside the container. To find the corresponding directory on the host VM, check the volume mounted at `$BACKUP_VOLUME`.

core/migrations/0037_auto_20240924_1020.py
```diff
@@ -0,0 +1,60 @@
+# Generated by Django 3.2.23 on 2024-09-24 08:20
```
vildead, line 1: Why is this migration needed? 

field `elu_accession`  was a field of CoreTrackedModel class (parent class of CoreTrackedDBModel which is parent class to all updated classes in this migration)

doc/backup.md
```diff
@@ -0,0 +1,122 @@
...
+- **Output**: `backup_<timestamp>.tar.gz` in the `BACKUP_DIR` (`../backups` by default).
+
+## Scheduled Backup with Cron
+
+To schedule the backup script to run automatically at a specific time using cron, add the following line to your crontab:
+
+1. Open the crontab editor:
+
+    ```bash
+    crontab -e
```
vildead, line 59: is this supposed to be system cron or cron inside the container?

doc/deployment.md
```diff
@@ -0,0 +1,164 @@
...
+
+3. Check if the cron job is added:
+
+    ```bash
+    docker compose exec backup crontab -l
+    ```
+
+Replace `/path/to/backup_script.sh` with the actual path to backup_script.
+
+### Restore Legacy Backup
```
vildead, line 142: @moustaphacheikh here we can include steps to update in settings.py file when restoring legacy backup

core/migrations/0037_auto_20240924_1020.py
moustaphacheikh: this migration was missing in the test instance. it was never pushed to the develop branch.

doc/backup.md
moustaphacheikh: yes the cron job is in the VM so it can stop and start docker-compose with no issue

doc/backup.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Enable Automatic Backups
+
+To ensure automatic backups are enabled, set `ENABLE_BACKUPS=true` (enabled by default):
+
+```bash
+ENABLE_BACKUPS=true docker compose up -d backup
+```
+
+This will configure automatic backups based on the `BACKUP_SCHEDULE`.
```
vildead, line 34: BACKUP_SCHEDULE is nowhere to be found

doc/backup.md
```diff
@@ -0,0 +1,116 @@
...
+  - `.:/code`
+
+### Operations
+
+#### Enable Automatic Backups
+
+To ensure automatic backups are enabled, set `ENABLE_BACKUPS=true` (enabled by default):
+
+```bash
+ENABLE_BACKUPS=true docker compose up -d backup
```
vildead, line 31: If a user already deployed the tool (`docker compose up -d`) this container is already running so starting it again is not needed.

Also, what is the purpose of having `backup` container running if we have already `scripts/backup_script.sh` being managed by OS level `crontab`?

vildead: or are there like 3 ways to run backup?
a) manually by backup_script.sh
b) automatic - call backup_script.sh by cron
c) automatic - `backup` container running it from withing the container based on some schedule?

scripts/backup_script.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Stop services that should not run during the backup
+docker compose stop nginx flower worker beat web mq solr
```
vildead, line 4: do we need to stop all the services?

doc/backup.md
moustaphacheikh: ENABLE_BACKUPS: This was part of a work-in-progress documentation. It was initially used when we had cron jobs running inside the container, but it's no longer in use.

moustaphacheikh: we need the `backup` container to have the necessary libraries to do the db backups. the VM does not have the database.

scripts/backup_script.sh
moustaphacheikh: to prevent any changes to the data while the backup is being created.

aleuk0 (approved):

scripts/backup_script.sh
vildead: The backup was running fine so far with the service running. Not that a short downtime is an issue.
We can discuss this with @neoflex  :)

PR #526, opened 2024-10-11
391 Rems external id

aleuk0:


core/models/access.py
```diff
@@ -185,6 +185,15 @@ def clean(self):
         help_text="Was the entry generated automatically, e.g. by REMS?",
     )
 
+    application_id = models.IntegerField(null=True, help_text="REMS application ID.")
```
moustaphacheikh, line 188: Are these mandatory?  blank =False will force the user to submit them when they create a new access instance

aleuk0: done

moustaphacheikh:
@aleuk0 It worked as expected. i tested the end-to-end process. great work

moustaphacheikh (approved):
@aleuk0 It worked as expected. i tested the end-to-end process. great work

Issue #536, opened 2025-01-15
Nginx sets 1MB limit on file uploads


vildead:
With the new docker compose deployment, settings from https://github.com/elixir-luxembourg/daisy/blob/develop/docker/nginx/nginx.conf  is used. 

By default, nginx sets `client_max_body_size` to 1MB. 

Change the template so it allows bigger files (20MB? 50MB?)

vildead:
Error message could be given to user if uploading too big file.

PR #551, opened 2025-02-20
increase client_max_body_size - nginx

neoflex:
default limit of 1MB is too low for the documents that need to be uploaded to daisy


aleuk0 (approved):

vildead:
Closes https://github.com/elixir-luxembourg/daisy/issues/536

PR #580, opened 2025-06-18
474 make contract box on project page more informative

moustaphacheikh:


core/forms/share.py
```diff
@@ -101,9 +100,6 @@ def __init__(self, *args, **kwargs):
                     "contract"
                 ].help_text = "No contract found with selected partner."
 
-            self.fields["contract"].choices = [
```
aleuk0, line 104: why should `self.fields["contract"].choices` be removed for `Logbook`?

moustaphacheikh: with the new name field, and the the __str__ method, we do not need to change the string representation here

core/models/contract.py
```diff
@@ -141,7 +149,18 @@ def partner_roles(self):
...
+        return self.name or self.short_name()
+
+    def display_partners(self):
+        partners = self.partners.all()
+        if len(partners) > 2:
+            partners_str = ", ".join(str(p) for p in partners[:2])
+            return f"{partners_str} and {len(partners)-2} more..."
+        return ", ".join(str(p) for p in partners) if partners else "-"
+
+    def display_partners_tooltip(self):
```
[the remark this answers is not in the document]
moustaphacheikh: done

core/forms/share.py
moustaphacheikh: done

aleuk0 (approved):

PR #589, opened 2025-08-28
v1.10.0

vildead:
release


vildead:
The migration was tested on stable instance - everything ok

PR #595, opened 2025-09-10
add restart policy to web container

neoflex:
this is meant to address an issue where keycloak bruteforce detection is triggered at web container startup


neoflex:
this is not enough apparently, we need a better way. maybe initialize the kc connection only when we need it, or at least catch the connection error and retry.

moustaphacheikh (approved):

PR #598, opened 2025-10-07
update settings and envs

### LINKED ISSUE #525: Reorganize settings.py files and move most of the variables into .env files
There are way too many settings.py files (ci, local, compose, ...) under `elixir_daisy` folder. 

- [ ] get rid of settings_local.py file
- [ ] get rid of the redundant settings_local files (maybe create several env files)
- [ ] rename `elixir_daisy` folder into something meaningful

aleuk0:


docker-compose.yaml
```diff
@@ -32,6 +34,8 @@ services:
     db:
         image: postgres:16.4-alpine
         restart: unless-stopped
+        ports:
+            - "5432:5432"
```
neoflex: any reason to map the port to the host machine? it's safer to have it only available inside the docker network and should be the default except if there is a good reason to do differently

docker-compose.yaml
```diff
@@ -78,6 +82,8 @@ services:
             interval: 10s
             timeout: 5s
             retries: 5
+        ports:
+            - "8983:8983"
```
neoflex: same comment here

neoflex:
no impact on documentation? (readme and other .md files)

docker-compose.yaml
aleuk0: I forgot to remove it for push, usually use it locally for development

conftest.py
moustaphacheikh, line 6: docker compose exec web pytest tests/test_models.py::test_user

`ImportError while loading conftest '/code/conftest.py'.
conftest.py:6: in <module>
    from django.contrib.auth.models import Group, Permission
/usr/local/lib/python3.12/site-packages/django/contrib/auth/models.py:5: in <module>
    from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
/usr/local/lib/python3.12/site-packages/django/contrib/auth/base_user.py:43: in <module>
    class AbstractBaseUser(models.Model):
/usr/local/lib/python3.12/site-packages/django/db/models/base.py:131: in __new__
    app_config = apps.get_containing_app_config(module)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/lib/python3.12/site-packages/django/apps/registry.py:260: in get_containing_app_config
    self.check_apps_ready()
/usr/local/lib/python3.12/site-packages/django/apps/registry.py:138: in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
E   django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.`

.env.test
```diff
@@ -0,0 +1,42 @@
+ENVIRONMENT=test
```
moustaphacheikh, line 1:
```
CELERY_TASK_ALWAYS_EAGER=True
CELERY_TASK_EAGER_PROPAGATES=True

And read them in settings.py:

CELERY_TASK_ALWAYS_EAGER = env.bool("CELERY_TASK_ALWAYS_EAGER", default=False)
CELERY_TASK_EAGER_PROPAGATES = env.bool("CELERY_TASK_EAGER_PROPAGATES", default=False)
```

conftest.py
aleuk0: Yes, you will get this error if you don't install the test dependencies:
`docker compose exec -T --user root web pip install ".[test]"`

.env.test
aleuk0: since we only use it for tests, I put it in `conftest.py` as `celery_config()`

moustaphacheikh (approved):
Great work !

PR #633, opened 2026-05-06
update default log level to INFO and remove debug filter from console…

### LINKED ISSUE #632: Improve logging for docker-compose deployment
**Is your feature request related to a problem? Please describe.**
docker compose deployment, when DEBUG is false the daisy logs only for to a file daisy.log inside the container. as a result, docker compose logs web is not really usable.
Also, it seems the default log level is ERROR for daisy.  

**Describe the solution requested**
daisy logs should flow to docker logs system.
would be nice to change the log_levels based on env variables or local settings.

aleuk0:
… handler

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -21,9 +21,6 @@ services:
     # Web Application
     web:
         <<: *app
-        environment:
-            - GUNICORN_LOGLEVEL=${GUNICORN_LOGLEVEL:-info}
-            - GUNICORN_ACCESS_LOG=${GUNICORN_ACCESS_LOG:-/dev/null}
         expose:
             - "5000"
         volumes:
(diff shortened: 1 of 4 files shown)


elixir_daisy/settings.py
```diff
@@ -234,7 +234,6 @@
     "handlers": {
         "console": {
             "level": "DEBUG",
```
neoflex, line 236: I would use LOG_LEVEL there instead of DEBUG.
But in general, I think we need to a more generic way to customize logging configuration. I think it's not flexible enough right now. If I have it running in docker compose, I may not want to have the daisy.log file inside the container for example.

aleuk0: but it is another layer, we have it in the "loggers", for the "handlers" we set the minimum threshold from which logs must be processed. should we duplicate this logic?

I would create a volume for logs files to persist. That's right, every time we create a new container we loose our logs.

if we want to send them to ELK, we can try this library:
https://github.com/django-es/django-elasticsearch-dsl
Or even better, we can use Filebeat to read from container stdout and send them to Logstash

neoflex: indeed, you are right as long as we have the log_level on the loggers level, it is ok. sorry, I missed that.
for the persistance, as discussed, let's use a volume and the file handler. We could still introduce two new env variables to toggle on or off the console and file handler as people might have different preferences for that. sounds ok?

docker-compose.yaml
```diff
@@ -35,7 +37,7 @@ services:
                 condition: service_healthy
             mq:
                 condition: service_healthy
-        command: sh -c "gunicorn -w 2 -b :5000 --pid /code/gunicorn.pid --log-level ${GUNICORN_LOGLEVEL:-info} elixir_daisy.wsgi"
+        command: sh -c "gunicorn -w 2 -b :5000 --pid /code/gunicorn.pid --log-level $$GUNICORN_LOGLEVEL elixir_daisy.wsgi"
```
neoflex: the double $$ character is intentional there?
also, could we have an option for the gunicorn access logs. It can be off by default but would be convenient to have a way to enable them ? See (https://gunicorn.org/reference/settings/#accesslog)

aleuk0: yes, we need the "$$" to use the param defined in line 25 "GUNICORN_LOGLEVEL", with just "$" it will use one from env

moustaphacheikh (approved):

PR #642, opened 2026-06-02
Redesign

moustaphacheikh:


web/templates/landing.html
```diff
@@ -0,0 +1,3981 @@
+{% extends 'layout.html' %}
```
aleuk0, line 1: this page should be removed from the "Redesign" branch, 3981 lines :)

web/views/dashboard.py
```diff
@@ -1,34 +1,304 @@
...
     context = {
         "last_datasets": last_datasets,
         "last_projects": last_projects,
-        # "notifications": notifications,
     }
-
     return render(request, "dashboard.html", context)
+
+
+def landing(request):
```
aleuk0, line 39: the same, landing

web/urls.py
```diff
@@ -120,6 +120,7 @@
 web_urls = [
     # Single pages
     path("", dashboard, name="dashboard"),
+    path("landing/", landing, name="landing"),
```
aleuk0, line 123: and here

aleuk0 (changes requested):

PR #643, opened 2026-06-19
develop to master 1.12.0

aleuk0:


scripts/update_app.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
vildead, line 1: What is this file used for? 

If the update is now automatized (good!), could you please update also the doc related to update/upgrade procedure?

Also, location of this file should be probably /scripts or core/management, no?

vildead (changes requested):
Are there any breaking changes with this one?  If so, could you please add section to https://github.com/elixir-luxembourg/daisy/blob/develop/doc/update.md#breaking-migrations ?

scripts/update_app.sh
aleuk0: yes, this script is now in /scripts, and [#breaking-migrations](https://github.com/elixir-luxembourg/daisy/blob/develop/doc/update.md#breaking-migrations) has also been updated

NimaGozalpour:
@vildead , this version has been tested on the stable and demo instances.

vildead (approved):

PR #647, opened 2026-07-07
fix release validation

vildead:
The workflow will validate the version in project.toml is updated upon release


vildead:
@aleuk0 the same here - very small change. Could you please approve so it can go to the next release? Thanks!

aleuk0 (approved):

PR #648, opened 2026-07-07
Styleguide

aleuk0:


vildead:
We have checked the changes on test instance ⭐ 

Few comments:

- The tabs are good but they should be more visually highlighted. Maybe making the text bigger or bolder... I will leave it to you :) (right now, it is just hard to spot them when being on an entitiy page for the first time)
- Could you please remove the Label/badge `Record complete`? It is not really defined what complete record looks like. You can probably keep the orange and red labels when something is obviously missing. 
- About page - the labels of sections (Acknowledgement and Resources) are very small. We suggest to make them blue bold (same as for example "Study details" on project page

vildead (changes requested):
☝🏼

vildead (approved):
great work

moustaphacheikh (approved):

